PR #39, opened 2018-06-04
MySql provider

andresrsanchez:
Hello! Various comments about this pull request:
-I don´t know how to get rid of 'System.Diagnostics.DiagnosticSource.dll', I have put this dll on VS IDE folder but this is not the solution I think. 
-AssemblyBindingRedirectHelper file is probably wrong, I don't understand well the dependencies management on this project.
-I strongly signed Pomelo.EntityFramework.MySql.dll with Key.snk on ReverseEngineer20 project.

Surely there are things wrong but I'm understanding the project and the code of MySql provider it's usable I think.

Thanks for all and sorry for my English.


ErikEJ:
Thanks for the pull request, am I correct in assuming that you would like to enable Pomelo MySql reverse engineering?

If that is correct, I think you are trying to many things at one time:

1: It looks like you are adding the official MySql provider to the reverse engineer project, not the Pomelo driver??

2: You are upgrading to 2.1 AND adding the MySql provider at the same time.

3: You should not touch the efpt project, it is not used for reverse engineering

4: You should not add app.config file, since this is an addin

5: You should not add .NET binaries to the lib folder

6: Make sure the few Unit tests pass before yo submit a PR

I suggest you close this, and wait for me to update the reverse engineering to 2.1, then you can make a pull request to add support for MySQL (I assume Pomelo?)

ErikEJ:
Dont be sorry, I apperciate your effort.

Pomelo uses the https://www.nuget.org/packages/MySqlConnector/ ADO.NET provider, so you can use that for MySqlConnection instead, I think

I will let you know when 2.1 has been enabled for reverse engineering!

ErikEJ:
@andresrsanchez Any interest in a PR for this?

PR #51, opened 2018-06-14
Added correct Feed URL

politigf:
The instructions for automatically getting the latest daily build are for SQLCeToolbox. Added correct Feed URL to use when following those instructions


ErikEJ:
Thanks!

Issue #14, opened 2018-05-19
investigate possibility for custom naming of dbset, classes and properties


ErikEJ:
_From @ErikEJ on February 22, 2018 9:10_

https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs 

https://github.com/aspnet/EntityFrameworkCore/issues/11056#issuecomment-368256127

- [x] Get PR for ICandidateNamingService accepted
- [ ] Add support for efpt.renaming.json (drop empty, commented out sample in project, in there is none)
- [ ] add class for renames ColumnRenamer: Table, Old, New & TableRenamer: Old, New
- [ ] Override ICandiadteNamingService, and rename based on json file

_Copied from original issue: ErikEJ/SqlCeToolbox#640_

ErikEJ:
_From @jonreis on March 14, 2018 19:41_

Erik, I really like your Reverse Engineering tool with Handlebars. Great work. Adding these items will be fantastic.

Do think you could add to the list specifying the datatype of the property also? As you know the generated code for Sqlite does not always reflect the datatypes the user wants to use.

ErikEJ:
@jonreis I can, but it is outside the scope of this particular feature, due to the way EF Core reverse engineering is designed.

ErikEJ:
_From @jonreis on March 14, 2018 19:58_

I guess I should download the code and see how everything works before I log requests. It seemed you would be able to change the type while you were renaming the property.

Thanks Erik.

ErikEJ:
No, I just thin you should log this as a seperate issue, the fix for that will no be possible via this feature/fix.

ErikEJ:
_From @statler on April 11, 2018 4:33_

Hi Erik

Any thoughts on when this may be available? Specifically the column renaming

ErikEJ:
@statler  Around the release of 2.1 (was waiting for changes in EF Core, which are now available in in preview-2) - what database provider do you use?

ErikEJ:
_From @statler on April 11, 2018 22:12_

Yes - I followed your links through and saw the commits - do you have a [non release] EFPT commit with the renaming in it, or still a work in progress?

I use full SQL server

ErikEJ:
Not yet, I need preview2, which was just released...

ErikEJ:
_From @OmiCron07 on May 7, 2018 21:40_

[...]

PR #87, opened 2018-08-25
Add Options and Replacement for table prefix

rposener:
Wanted to submit this back to your awesome code base.  This is making our migration from EF6 => EfCore so much simpler!

I added a feature here that allows replacement of table prefixes.  The same could probably be done for suffix or other options as well. 

We used to use @sjh37's [ReversePoco.com](https://github.com/sjh37/EntityFramework-Reverse-POCO-Code-First-Generator) project for our EF6 models.  One of the features we really liked was the ability to remove the tbl bit from all of the tables in our legacy database without affecting the project.

Thanks - if there is more I can do to help get this pulled into your next release, let me know.


ErikEJ:
Closing, will be done in #14

PR #108, opened 2018-10-17
invert property naming for fk.

tomzre:
This is the proposal to solve problem with wrong Inverse Property names.

### DIFF
--- a/src/GUI/ReverseEngineer20/ReverseEngineer/ReplacingCandidateNamingService.cs
+++ b/src/GUI/ReverseEngineer20/ReverseEngineer/ReplacingCandidateNamingService.cs
@@ -1,4 +1,6 @@
-﻿using Microsoft.EntityFrameworkCore.Scaffolding.Internal;
+﻿using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Scaffolding.Internal;
 using Microsoft.EntityFrameworkCore.Scaffolding.Metadata;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


tomzre:
Can you let me know is this change can be merged?

ErikEJ:
I would like to know more about the effect of using it - can you show me a sample DbContext ?

tomzre:
In general this is about FK to table eg.

```
DBContext
public virtual DbSet<MasterEmployee> MasterEmployee { get; set; }
public virtual DbSet<MasterCustomer> MasterCustomer { get; set; }
public virtual DbSet<MasterCompany> MasterCompany { get; set; }
```
And if in `MasterEmployee `we have a FK to Customer
we have a property
```
[ForeignKey("CustomerId")]
[InverseProperty("Employee")]
public MasterCustomer Customer { get; set; }
```
And `MasterCustomer` has following property for this FK
```
[InverseProperty("Customer ")]
public ICollection<MasterEmployee> MasterEmployee{ get; set; }
```
The `Master` is the schema name which I gave for example of custom name.

As you can see in the InverseProperty Schema name is missing. 
Please let me know if you want to more information about that.
Thanks

ErikEJ:
OK, would it be possible to add a test?

tomzre:
I am going to try to add some. But in the meanwhile I found the self referencing key changes the name to
InverseName and InverseNameNavigation. I am going to investigate it further.

tomzre:
Ok, now I know how the Inverse and Navigation is added. This is desired behaviour for Unique property names in the class and it's happening in the pipeline after the column name goes through the overridden service. I added some tests and actually tested it on production and it seems to be ok. Please review and let me know if you need more information.

ErikEJ:
Thanks!

tomzre:
Is this version available from vs studio marketplace?

ErikEJ:
No, as a daily build only for now

Issue #112, opened 2018-10-24
[Feature Request] List tables without primary key


Herdo:
Tables without a primary key don't show up in the `Select Tables to Script` dialog.
There's no further reason specified why the table isn't in the dialog, even though the database projects builds fine.

Suggestion: List the tables without primary key in the dialoge, but make the checkbox unchecked and disable the control, with a note showing behind the table that the primary key is missing.

### Steps to reproduce
Example: [MissingTableConsole.zip](https://github.com/ErikEJ/EFCorePowerTools/files/2511172/MissingTableConsole.zip)

1. Open attached solution
2. Right-click on the `MissingTableConsole` project
3. `EF Core Power Tools` --> `Reverse Engineer`
4. Select `MissingTableDatabase` in the DacPac drop-down
5. Table `config.Message` is not showing up

### Further technical details
EF Core Power Tools version: 1.0.664.0

Database engine: None (DacPac)

Visual Studio version: Visual Studio 2017 15.8.7

Herdo:
@ErikEJ I would pick this, if you're okay with the suggestion.

On that topic, in `DacpacTableListBuilder.cs`, is there a reason/case why the check is for `.Where(t => t.PrimaryKeyConstraints.Count() > 0)` instead of `== 1`?

ErikEJ:
>> I would pick this, if you're okay with the suggestion.
Sounds fantastic! Keep in mind, that ideally we should do the same for SQL Compact and NpgSQL (and SQL Server databases)

>> why the check is for .Where(t => t.PrimaryKeyConstraints.Count() > 0) instead of == 1?

Maybe in order to accomodate multi column primary keys?

caojinqin:
When I am using reverse engineering of EF Core Power Tools, I often report an error. The Core version I installed is Core2.1.

ErikEJ:
@caojinqin Which error??

Herdo:
No, the changes in #114 
I changed the `ReverseEngineerOptions.Tables` from `List<string>` to `List<TableInformation>` in that PR. So the configuration file saved with the old `List<string>` cannot be deserialized with the new version of that class.

I just found this issue while manually testing #124

ErikEJ:
I see. Breaking change. Maybe add a new "TableInfos" property, and revert Tables to be List<string>

ErikEJ:
Are the changes to DatabaseType in #124 backwards compatible?

ErikEJ:
[...]

PR #114, opened 2018-10-25
Implemention for #112

Herdo:
Implementation for #112 

UI could use some additional work, but with the current UI framework, this is the easiest solution.

Regarding other providers besides DacPacs:
I included the information for existing primary keys where the information is already present.
Other providers needs to be extended, I lack the knowledge for that.


ErikEJ:
Remember to add tests and update existing

Herdo:
Dialog changes:
grafik

Tooltip on icon:
grafik

ErikEJ:
Looks good (under the circumstances, as you also point out)

Issue #116, opened 2018-10-26
Refactor dialogs to implement MVVM pattern


Herdo:
The current dialogs in the `EFCorePowerTools` project include business logic in their background classes. This makes the dialogs not really testable, as business logic requires an instance of the dialog to be created.

Implement the MVVM pattern for the dialogs to seperate the UI logic from the business logic.

### Requires:
- New shared project required (probably .NET Standard 2.0), e.g. `EFCorePowerTools.Shared`
- Adding dependency injection
- Loose coupling between model (data access), view model (view logic, model connector) and view (dialog only)

### Benefits:
- Components (dialogs and other layers) will be testable/better testable
- Stability through more tests
- Easier extensibility through seperate layers

Herdo:
@ErikEJ I refactored the "About" dialog, which ends up in a very similiar looking dialog:
**Current:**
![grafik](https://user-images.githubusercontent.com/651191/47607341-ca238680-da1e-11e8-9b51-c3023f4d542d.png)

**Refactored:**
![grafik](https://user-images.githubusercontent.com/651191/47607346-d4458500-da1e-11e8-9c35-4969c7ebbef7.png)

Feel free to take a look at the commit: https://github.com/Herdo/EFCorePowerTools/commit/0d7ca289ac26cb47d0a3ad1d854e292a52dc8366 Of cause, there's some ground work in there, that'll be reused for other dialogs.

Yet, the refactoring to MVVM results in smaller and testable units. 👍 
Of cause, the refactored version is fully functional.

I would go for the remaining dialogs, if you like the changes.

ErikEJ:
This is excellent, thanks a lot - and I really like small, targeted PRs!

ErikEJ:
I had a look, and really like the changes, please keep future PRs as small as possible (understand the size because of groundwork here)

Herdo:
Would you be fine with one PR for each dialog?
I think it won't make much sense making too small ones, as main probably won't build then...

I can, however, try to make small commits inside each pull request. If that's fine by you.
I guess that's still easier to review.

I'd add multiple PRs after changing all dialogs to get the logic covered with tests.

ErikEJ:
A PR per dialog would be just fine 😎👍

Herdo:
@ErikEJ Currently refactoring the `PickServerDatabaseDialog`. Would you have any problems with me renaming it?

Proposal: `ReverseEngineerSourceSelectionDialog`.
[...]

PR #117, opened 2018-10-28
Refactored "About" dialog

Herdo:
I had a small fight with the rebase, but I hope I got it right now with this PR.
You might want to double check 😄 

Summary of commit 1:
- Added 2 new projects (**EFCorePowerTools.Shared** and **EFCorePowerTools.BLL**)
  - Had to add 2 key files for those projects, otherwise VS doesn't allow debugging (might want to review this for publishing)
  - Shared holds interfaces and models
  - BLL holds business logic
- Added ViewModels and data access layer in the main project
  - ViewModels and business logic can now be tested without any reference to hard dependencies, rather the tests can focus on the interfaces
  - Wrapped some static helper classes in Interface implementations, so they can be mocked in unit tests
- Basic implementation for MVVM
  - Added dependency injection to the "Package" class
  - Can resolve a view by interface
  - Added message handler in the "Package" class to handle message boxes
  - Had to add a Resources dictionary to the shared project for some guids (required in main and BLL project)
- Refactored `About` dialog to work with the MVVM pattern
  - Replaced events and replaced them with commands
  - Adjusted XAML to use static resources and bindings

Summary of commit 2:
- UnitTests failed due to version mismatch of `System.ValueTuple`, therefore I added the NuGet reference


Herdo:
Part 1 for #116

src/GUI/EFCorePowerTools/app.config
```diff
@@ -0,0 +1,31 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
ErikEJ: Remove this file

ErikEJ:
Thanks a lot, great contribution!

PR #124, opened 2018-11-04
Refactored PickServerDatabase dialog

Herdo:
For #116 

#### Commit 1 (1b77674)
- Added database connection model to hold information about a database connection
- Added database definition model to hold information about a database definition (database project or dacpac)
- Added unit tests for the new models
- Added shared enum DatabaseType

#### Commit 2 (3280b27)
- Replaced the DatabaseType enum in `EFCorePowerTools` with the shared enum from commit 1
- Synced enum in ReverseEngineer (**Note: there's always the risk that those enums won't match at any point in time, if somebody forgets to update both**)

#### Commit 3 (9e1f6ee)
- Implemented ViewModel for the PickServerDatabase dialog
- Added unit tests for the new ViewModel

#### Commit 4 (86758b8)
- Refactored the actual dialog (*.xaml and *.xaml.cs) by implementing the view model

### DIFF
--- a/src/GUI/EFCorePowerTools/Dialogs/PickServerDatabaseDialog.xaml.cs
+++ b/src/GUI/EFCorePowerTools/Dialogs/PickServerDatabaseDialog.xaml.cs
@@ -1,99 +1,74 @@
-﻿using System;
-using System.Windows;
-using System.Collections.Generic;
-using ErikEJ.SqlCeToolbox.Helpers;
-
-namespace ErikEJ.SqlCeToolbox.Dialogs
+﻿namespace EFCorePowerTools.Dialogs
 {
-    using EFCorePowerTools.Shared.Enums;
+    using System;
+    using System.Collections.Generic;
+    using System.Windows;
+    using Contracts.ViewModels;
+    using Contracts.Views;
+    using Shared.DAL;
+    using Shared.Models;
 
-    public partial class PickServerDatabaseDialog
+    public partial class PickServerDatabaseDialog : IPickServerDatabaseDialog
     {
-        public KeyValuePair<string, DatabaseInfo> SelectedDatabase { get; private set; }
+        private readonly Func<(DatabaseConnectionModel Connection, DatabaseDefinitionModel Definition)> _getDialogResult;
+        private readonly Action<IEnumerable<DatabaseConnectionModel>> _addConnections;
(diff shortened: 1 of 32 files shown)


Herdo:
@ErikEJ I don't know why git thinks it should be in this PR, but `2c2ce2c` would be a fix for the issue mentioned in #112 - it'd be a migration route.

ErikEJ:
Did you forget to change branch?

Herdo:
Yeah, was still on master.

ErikEJ:
I usually create a separate branch in my clone for work on PRs

ErikEJ:
Have you done a manual test of the config file migration?

Herdo:
Yes, I just tested it.
Old version will be correctly migrated on read, with all checkboxes ticked, though, because the information about the existing primary key wasn't persisted in V1.
Upon saving, the new version was saved, including the new information.
After doing the process again, now with the new version, reading is also possible, however, the table without primary key is not ticked any longer (as it doesn't have a primary key).

Looking good so far.

ErikEJ:
Sounds great, I will review soon so we can get it in the daily build. Really appreciate all your contributions

ErikEJ:
Looks really good, just a couple of small comments

src/GUI/EFCorePowerTools/Dialogs/PickServerDatabaseDialog.xaml.cs
```diff
@@ -1,97 +1,74 @@
...
+            DatabaseConnectionCombobox.Focus();
         }
 
-        private void AddButton_OnClick(object sender, RoutedEventArgs e)
+        public (bool ClosedByOK, (DatabaseConnectionModel Connection, DatabaseDefinitionModel Definition) Payload) ShowAndAwaitUserResponse(bool modal)
         {
-            try
-            {
-                var info = EnvDteHelper.PromptForInfo(_package);
-                if (info.DatabaseType == DatabaseType.SQLCE35) return;
```
Herdo: @ErikEJ Regarding the return unknown, do you mean this?
The old cold just returned, if it's `SQLCE35`.

If not, please explain a little bit more what you mean by "return unknown" on the new implementation 😄

ErikEJ: Make PromptForInfo return DatabasteType.Unknown instead of SQLCE35

PR #130, opened 2018-11-22
Added additional null-checks, just in case

Herdo:
For the bug reported in #116 
Should add a little bit more safety.


ErikEJ:
Thanks!

PR #139, opened 2018-11-29
More tests for the PickTablesViewModel

Herdo:
- Added more tests for the ViewModel
- Moved more logic from the Dialog background class into the ViewModel


ErikEJ:
Thanks!!

PR #146, opened 2018-12-11
Completely disabled selection for tables without primary key

Herdo:


ErikEJ:
This only works for reverse engineering from a dacpac currently, is that correct?

roji:
@ErikEJ you're asking about how to check if a table has a primary key, right? I don't really see anything like this in [the SqlServer schema collections](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-schema-collections#columns), how do you currently accomplish this for SqlServer?

Note that in `DbDataReader.GetSchemaTable()` [there's an `IsKey` property(https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqldatareader.getschematable?redirectedfrom=MSDN&view=netframework-4.7.2#System_Data_SqlClient_SqlDataReader_GetSchemaTable) which should also be well-supported by Npgsql, but that doesn't seem to exist in the data returned by `DbConnection.GetSchema()`...

ErikEJ:
I run a custom query!
https://github.com/ErikEJ/SqlCeToolbox/blob/master/src/API/Repositories/ServerDBRepository.cs#L491

Issue #166, opened 2019-01-15
Support Handlebars Helper and Transform Functions


tonysneed:
In general it would be good to support extensibility of the code generation process by allowing registration of services via `IServiceCollection`.  This is needed to support adding helper and transform functions, but it could also be useful for replacing other registered services, `ICSharpDbContextGenerator` or `ICSharpEntityTypeGenerator`.  I would suggest using the same mechanism as the EF Core CLI tooling, which is to look for a class that implements `IDesignTimeServices` and call `ConfigureDesignTimeServices`, passing the `IServiceCollection` created in `EfCoreReverseEngineer`.

This would allow devs to add something like the following to the project to register a Handlebars helper function.

```csharp
public class ScaffoldingDesignTimeServices : IDesignTimeServices
{
    public void ConfigureDesignTimeServices(IServiceCollection serviceCollection)
    {
        var myHelper = (helperName: "my-helper", helperFunction: (Action<TextWriter, Dictionary<string, object>, object[]>) MyHbsHelper);

        // Add optional Handlebars helpers
        serviceCollection.AddHandlebarsHelpers(myHelper);
    }

    // Sample Handlebars helper
    void MyHbsHelper(TextWriter writer, Dictionary<string, object> context, object[] parameters)
    {
        writer.Write("// My Handlebars Helper");
    }
}
```

Relates to #16, #126.

tonysneed:
@ErikEJ If you agree with this approach, I would be happy to submit a PR to search for and use an `IDesignTimeServices` implementation.

ErikEJ:
EF Core Power Tools runs from VS, completely outside any user code. So I cannot see how this will ever be invokde?

tonysneed:
EF core CLI has a similar problem. It should be possible via reflection or late bound invocation.

tonysneed:
Would you consider a PR?

ErikEJ:
You will be entering a world of pain ,but go ahead. I cannot see how you can load types without compiling, and then you have the joy of non present dependencies and .net core etc

tonysneed:
At first I was thinking of non-compilation for a simple scenario of adding Handlebars helpers in a single file. But I’m thinking now this is not such a good idea.

Full extensibility is really what is needed, and that requires compiling, which of course is what the EF Core CLI does. If its OK, let me experiment a bit with this.

ErikEJ:
Should we close this?

tonysneed:
[...]

PR #170, opened 2019-01-18
Add DesignTimeServicesBuilder for configuring design-time services

tonysneed:
Relates to issue #166.


tonysneed:
There are currently a few issues with this solution.
- EF Core PT need to run on a NET Standard library (not .NET Core) where a `IDesignTimeServices` class is present.
- Project needs to reference the same version of Microsoft.EntityFrameworkCore.* that the EFCorePowerTools reference.
- `Assembly.LoadFrom` locks the bin folder of the project so it can't be re-built.
- The `IDesignTimeServices` is unloadable if it references services in other projects.

ErikEJ:
Given the number of issues and restrictions, I am concerned about the value and supportability of this feature.

tonysneed:
@ErikEJ Yes, I agree. It was worth the effort to investigate the feasibility of the feature, but the issues involved make it difficult to implement in reliable fashion.

PR #175, opened 2019-02-03
Refactored EfCoreModelDialog to MVVM

Herdo:
Changes for #116 

Before:
grafik

After:
grafik

@ErikEJ This would require some additional QS from your side. I think I got all the logic right as it was, but a second look might not hurt 😉 
Also I had a "how could this ever work" moment in `EfCoreReverseEngineer.GenerateFiles`.


ErikEJ:
Thanks, Stefan, I will take a look soon, just waiting for the 2.2 release dust to settle.

ErikEJ:
Looks good and appears to work as desired (I did some smoke testing) - only comment is the spacing between the check boxes in the Naming section

Herdo:
@ErikEJ Less squishy:
grafik

PR #183, opened 2019-02-18
Bugfix for pick server / database dialog

Herdo:
Adding a new connection via the 'Add...' wouldn't display anything in the ComboBox after adding a new conneciton. There would be an entry, but without any text. Those entries were however displayed correctly when opening the dialog again.

This fix resolves the issue, by keeping the display name throughout the layers.


ErikEJ:
Perfect, spotted that myself recently, but forgot to file an issue

Issue #206, opened 2019-04-21
Pre-select database project when reverse engineering


Herdo:
When a solution contains a database project (*.sqlproj), and there's no database connection currently registered, the database project should be pre-selected as default.

### Further technical details
- If there's a database test project (file name to lower contains "test"), this shall not be pre-selected.
- If there's no database project, but only DACPAC files, those should neither be selected.

PR #207, opened 2019-04-21
Pre-select database project if no database connection is available

Herdo:
Implementation for #206

### DIFF
--- a/src/GUI/EFCorePowerTools/ViewModels/PickServerDatabaseViewModel.cs
+++ b/src/GUI/EFCorePowerTools/ViewModels/PickServerDatabaseViewModel.cs
@@ -123,18 +123,7 @@
 
         private DatabaseDefinitionModel PreSelectDatabaseDefinition()
         {
-            var subset = DatabaseDefinitions.Where(m =>
-                                             {
-                                                 if (string.IsNullOrWhiteSpace(m.FilePath)
-                                                  || !m.FilePath.EndsWith(".sqlproj"))
-                                                 {
-                                                     return false;
-                                                 }
-
-                                                 var fileName = Path.GetFileNameWithoutExtension(m.FilePath).ToLower();
-                                                 return !fileName.StartsWith("test")
-                                                     && !fileName.EndsWith("test");
-                                             })
+            var subset = DatabaseDefinitions.Where(m => !string.IsNullOrWhiteSpace(m.FilePath) && m.FilePath.EndsWith(".sqlproj"))
                                             .ToArray();
             return subset.Any()
                        ? subset.OrderBy(m => Path.GetFileNameWithoutExtension(m.FilePath)).First()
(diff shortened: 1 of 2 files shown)


src/GUI/EFCorePowerTools/ViewModels/PickServerDatabaseViewModel.cs
```diff
@@ -109,5 +120,24 @@ private void Cancel_Executed()
...
+            var subset = DatabaseDefinitions.Where(m =>
+                                             {
+                                                 if (!m.FilePath.EndsWith(".sqlproj"))
+                                                 {
+                                                     return false;
+                                                 }
+
+                                                 var fileName = Path.GetFileNameWithoutExtension(m.FilePath).ToLower();
+                                                 return !fileName.StartsWith("test")
+                                                     && !fileName.EndsWith("test");
```
ErikEJ: This seems brittle?

src/GUI/EFCorePowerTools/ViewModels/PickServerDatabaseViewModel.cs
```diff
@@ -109,5 +120,24 @@ private void Cancel_Executed()
             SelectedDatabaseDefinition = null;
             CloseRequested?.Invoke(this, new CloseRequestedEventArgs(false));
         }
+
+        private DatabaseDefinitionModel PreSelectDatabaseDefinition()
```
ErikEJ: Possible to add a test?

Herdo: Will do.

src/GUI/EFCorePowerTools/ViewModels/PickServerDatabaseViewModel.cs
Herdo: To what extend?
There's no better way to find "test" projects, rather than by naming convention. 🤔

ErikEJ: Are these test database projects? (sqlproj) ?

ErikEJ: I think this must be removed, there is only .sqlproj

Herdo: As said, it's by naming convention, not by project type or file extension.

e.g. you have these two database projects:

- BookingDatabase.sqlproj
- BookingDatabase.Test.sqlproj

`BookingDatabase.sqlproj` is the primary database project, defining the schema.
`BookingDatabase.Test.sqlproj` references `BookingDatabase.sqlproj`, and is used for test-deployments (e.g. builds) of the main project, seeding data, etc etc.

That's what the filter is for. And it doesn't filter the available projects in the dropdown, it just filters for the pre-selection. So if you have those two projects from the example in your solution, it'll pre-select the primary database project, rather than a project that doesn't define any objects in itself.

ErikEJ: I still think this is too localized and not a standard convention.

Issue #197, opened 2019-03-28
Reverse Engineer - Ability to have seperate folders for DataContext and Models in same project


Moorepheus:
Would be nice to have separate inputs for the Data Context and Model path.

Moorepheus:
Would be nice to have separate entries for the DataContext and Model path.

ErikEJ:
You can run the code gen twice, and then specify the folder/project for each run (and select to only generate Models or DbContext)

ErikEJ:
Ping? Are you unblocked?

Moorepheus:
Sorry, was out of the office. Yes, I understand you can run it twice, but that is an extra step that can be avoided.  Also, too many mistakes can be done having to run it twice.  Thanks for the help anyway.

ErikEJ:
Reopening to enable upvotes

tyrinj1120:
This would be a really nice feature. I was actually looking around the ui for it. Great tools Erik.

ErikEJ:
@tyrinj1120 You can do this today by running the process twice, did you try that?

Moorepheus:
> @tyrinj1120 You can do this today by running the process twice, did you try that?

Yes, but this leaves the door wide open for mistakes.  While I can do this all in one command line, this tool saves on having to type it out what I want to do, and if the ability was to have this feature in this product would also make it more efficient.

ErikEJ:
@Moorepheus So you would like to text boxes on the form, by default same path in both?

Moorepheus:
Yes, one text box for the Data Context Path and one text box for the Models path. They could be the same or different, depending on what the user wants to do. In our case, they are different. , This would make it more flexible then having to run it twice just because we want different paths for the Data Context and the Models.

ErikEJ:
>> , by default same path in both?

Thansk, this was a question about the initial values!

Moorepheus:
I think the initial values could be the same, I don't see how that would be an issue.

jim-paton:
Is there any news on this? Awaiting this feature is currently blocking our adoption of the extension.

ErikEJ:
Why are you blocked? https://github.com/ErikEJ/EFCorePowerTools/issues/197#issuecomment-477608575

jim-paton:
[...]

PR #229, opened 2019-06-22
Add #197 - 2 OutputPath (Models and Context)

Wasseratem:


ErikEJ:
Build failed...

Wasseratem:
Well, I see that, but I can't see why. It is working on my machine and I can't see an reason on the link at "appveyor" (what ever this tool is)

ErikEJ:
You forgot to check in a file, I think:
CSC : error CS2001: Source file 'C:\projects\efcorepowertools\src\GUI\ReverseEngineer20\ReverseEngineer\OutputPath.cs' could not be found. [C:\projects\efcorepowertools\src\GUI\ReverseEngineer20\ReverseEngineer20.csproj]

Also, (re missing design discussion) - I propose that we stay backwards compatible, and just add a textbox below the current output path, for an optional path for the DbContext. If nothing is entered, just use the old logic, if entered, override the DbContext path.

src/GUI/EFCorePowerTools.Shared/Models/ModelingOptionsModel.cs
```diff
@@ -17,7 +17,8 @@ public class ModelingOptionsModel : INotifyPropertyChanged
         private bool _usePluralizer;
         private bool _useDatabaseNames;
         private string _ns;
-        private string _outputPath;
```
ErikEJ: Keep this to stay backwards compatible, and just add new optional path for DbContext

src/GUI/EFCorePowerTools.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2003
+# Visual Studio Version 16
```
ErikEJ: Revert this

src/GUI/EFCorePowerTools/Extensions/ReverseEngineerOptionsExtensions.cs
```diff
@@ -24,6 +24,10 @@ public static ReverseEngineerOptions TryRead(string optionsPath)
             if (couldRead)
                 return deserialized;
 
+            couldRead = TryRead(optionsPath, out ReverseEngineerOptionsV2 deserializedV2);
```
ErikEJ: Stay backwards compatible

src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
```diff
@@ -183,7 +184,11 @@ public async void ReverseEngineerCodeFirst(Project project)
                     ContextClassName = modelingOptionsResult.Payload.ModelName,
                     DatabaseType = (ReverseEngineer20.DatabaseType)dbInfo.DatabaseType,
                     ProjectPath = projectPath,
-                    OutputPath = modelingOptionsResult.Payload.OutputPath,
+                    OutputPath = new OutputPath()
```
ErikEJ: Just add a single new optioanl property

src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
```diff
@@ -271,6 +276,7 @@ public async void ReverseEngineerCodeFirst(Project project)
                     _package.LogError(revEngResult.EntityWarnings, null);
                 }
                 Telemetry.TrackEvent("PowerTools.ReverseEngineer");
+                throw new Exception("Test");
```
ErikEJ: Remove this line

src/GUI/ReverseEngineer20/EFCoreReverseEngineer.cs
```diff
@@ -125,7 +122,7 @@ public EfCoreReverseEngineerResult GenerateFiles(ReverseEngineerOptions reverseE
                         : reverseEngineerOptions.ConnectionString,
                     reverseEngineerOptions.Tables.Select(m => m.Name).ToArray(),
                     schemas,
-                    @namespace,
+                    modelNamespace,
                     "C#",
                     null,
```
ErikEJ: You can use this for the path to the context!

Wasseratem:
Hi Erik,

you are correct with the throw new Exception and ofc the vs 16 lines,
But what I don't understand is the second variable.
Your argument is that we should stay backwards compatible. With the new version of the configuration I am backwards compatible. I had the idea of "just adding a second variable, like OutputContext" too - but in my opinion you should never be "scared" to rename a variable to have a better understanding for everyone on your code. With "OutputPath" and "OutputPathContext" we may ask ourself sometimes "which is which".

ErikEJ:
The issue is that thousands of users already have saved config.json files with the existing variable name.

Wasseratem:
This isnt an issue.
As I told you, I developed a version of the config to merge this. That is the reason I made OutputPath a Class, Not 2 new strings (and ignore the old one).
When the JSON deserializer will come to that part with OutputPath it can't serialize it, because its a string and he is now awaiting an object.

See here:

`  var couldRead = TryRead(optionsPath, out ReverseEngineerOptions deserialized);
            if (couldRead)
                return deserialized;

            couldRead = TryRead(optionsPath, out ReverseEngineerOptionsV2 deserializedV2);
            if (couldRead)
                return ReverseEngineerOptions.FromV2(deserializedV2);

            couldRead = TryRead(optionsPath, out ReverseEngineerOptionsV1 deserializedV1);
            if (couldRead)
                return ReverseEngineerOptions.FromV1(deserializedV1);`

and

`  OutputPath = new OutputPath()
                {
                    Models = v2.OutputPath,
                    Context = v2.OutputPath,
                },`

This will cause the new OutputPath.Models and OutputPath.Context to the same value as the string OutputPath.
It's completly downgradeable without problems :)

Wasseratem:
Hi Erik,

any update on my argumentation? You want me to build it your way?

src/GUI/ReverseEngineer20/EFCoreReverseEngineer.cs
```diff
@@ -134,30 +131,57 @@ public EfCoreReverseEngineerResult GenerateFiles(ReverseEngineerOptions reverseE
...
+            var @namespace = root;
+            if (!string.IsNullOrEmpty(path))
+            {
+                @namespace += "." + path.Replace(Path.DirectorySeparatorChar, '.').Replace(Path.AltDirectorySeparatorChar, '.');
+            }
+
+            return @namespace;
+        }
+
+        private string PostProcessContext(string contextFile, ReverseEngineerOptions options, string contextNamespace, string modelNamespace)
```
ErikEJ: No need for this, just use the DbContextPath parameter mentioned above

ErikEJ:
I am unsure, but a big fan of as little code and as small changes as possible. My proposal will enable that, I believe. Also, there are other unresolved comments.

Wasseratem:
Well, yes, you made your point clear :) But sadly, that is not what I think what software development should be. At the moment a dev is too scared to have changes in his code, its most likly a sign of a problem with the codebase, e.g. too less tests etc. I think we should never be afraid of making changes that will help us for the rest of our life (or the life of our software, whatever comes first). 
I see that maybe I am not the right contributor for your big project :) Feel free to accept or adopt my changes of course!

ErikEJ:
Thanks for the effort - btw I do not see any additional tests in this PR...

Issue #268, opened 2019-10-06
Remaining EF Core 3.0 rev engineer features


ErikEJ:
Add missing EF Core 2 features:

- [x] Dacpac support
- [x] MySQL support
- [x] Custom renaming support
- [x] Test SQLite
- [x] Visualize keyless tables and views

ErikEJ:
@jespersh Looks like MySQL provider is now RC - should I wait for RTM?

jespersh:
I only just now noticed. 
Looking at https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/milestone/9 then it should be possible to use RC1.

Give me a poke if I should look into upgrading the packages. 
Will upgrading the packages prevent 2.x libraries from working?

ErikEJ:
I will have a look at adding the RC provider. The EF Core 3 scaffolding runs completely out of process, so no conflict.

ErikEJ:
@Herdo Do you have time to help, I need two additional text boxes with labels (below the 3 existing one) on this dialog - just need help with the XAML layout, I can do the viewModel /Databinding myself - TIA!


![efcreveng2](https://user-images.githubusercontent.com/4169187/66412214-63ac3400-e9f5-11e9-8630-176de90746ef.png)

jespersh:
@ErikEJ https://github.com/ErikEJ/EFCorePowerTools/commit/8326d8f2d97d48798904c598c204633baa4c3b3a that's all you had to add? cool! Will look forward to testing.

ErikEJ:
Yes, that's all. EF Core 3.0 reverse engineering is executed by a command line tool!

Herdo:
@ErikEJ What should I use as upstream branch for this?
Want me to put the correct labels there? Whatever those are? 😄

ErikEJ:
@herdo master - just add some dummy labels 😄

ErikEJ:
And thanks :grateful:

Alex-Torres:
What about null-able reference types? Is there a way to update the template engine to decorate class properties with ? if database fields are null?

ErikEJ:
@Alex-Torres This issue is for feature parity with the EF Core 2 scaffolding.

If you have feature suggestions, please crate a new issue, with sample code that demonstrates the changes you are asking for.

Alex-Torres:
[...]

PR #272, opened 2019-10-08
Extended dialog "Generate EF Core Model"

Herdo:
For #268 - here's the result:
extended dialog


ErikEJ:
Wauw!! That was quick!!

Issue #308, opened 2019-11-20
Pluralizer issue


qazq:
I have an original DbContext that generated by EF core power tool (I'm not sure the exact version, maybe is v2.2.12)

One part of my DbContext file is below:
```cs
public virtual DbSet<AndroidOta> AndroidOtas { get; set; }
public virtual DbSet<WorkBarrelAlias> WorkBarrelAliases { get; set; }
```
I added a new table to the SQL Server, so I re-generated the lastest DbContext by EF core power tool v2.3.86.0. 

But the original `AndroidOta` is changed

```cs
public virtual DbSet<AndroidOtum> AndroidOta { get; set; }
```

I set `"UseLegacyPluralizer": true` and generate again, but the `WorkBarrelAlias` is changed.

```cs
public virtual DbSet<WorkBarrelAlia> WorkBarrelAlias { get; set; }
```

`Alias` is a word, but the result is `Alia`. The 's' is removed.

I try to use `efpt.renaming.json` to resolve it, but it does not work.

```json
[
  {
    "UseSchemaName": false,
    "SchemaName": "dbo",
    "Tables": [
      {
        "Name": "work_barrel_alias",
        "NewName": "WorkBarrelAlias"
      }
    ]
  }
]
```

It seems to do nothing when the `Name` equal to `NewName`. If I change the `NewName` to `WorkBarrelAlias2`, the result is

```cs
public virtual DbSet<WorkBarrelAlias2> WorkBarrelAlias2 { get; set; } // WorkBarrelAlias2  should be WorkBarrelAlias2s ?
```
My question is, how can I force using the new table name and the tool generate pluralize variable name. Just like old version

```cs
public virtual DbSet<AndroidOta> AndroidOtas { get; set; }
public virtual DbSet<WorkBarrelAlias> WorkBarrelAliases { get; set; }
```

Thanks a lot.

### Steps to reproduce
Include a complete description that I can redo to reproduce the issue.

### Further technical details
EF Core Power Tools version: 2.3.86.0

Database engine: SQL Server

Visual Studio version: Visual Studio 2017 15.9.5

ErikEJ:
I think it is not currently possible

PR #316, opened 2019-11-26
Enhance customer rename

Hsu-Cheng-Wei:
Hello! This pr can solve this question https://github.com/ErikEJ/EFCorePowerTools/issues/308

I override EntityFrameworkCore RelationalScaffoldingModelFactory.cs to achieve


ErikEJ:
Thanks, that is a lot of code. 

Maybe a better approach would be to allow customization of the pluralizer via a configuration file?

Hsu-Cheng-Wei:
I think another way is modify pluralizer extension by self or wait  EntityFrameworkCore release better function

PR #344, opened 2020-02-02
Move SqlServerDacpacDatabaseModelFactory to its own package

bricelam:
This will enable the following command:

``` sh
dotnet ef dbcontext scaffold ..\MyDatabase\bin\Debug\MyDatabase.dacpac ErikEJ.EntityFrameworkCore.SqlServer.Dacpac
```

I'm not sure the resulting nupkg includes all the right dependencies. I'll test it out and update it when I get a chance.


bricelam:
Well, it runs now but it's not finding any tables. Probably something wrong with my dacpac.

ErikEJ:
I wonder if it related to this LINQ filter: https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/ReverseEngineer20/ReverseEngineer/SqlServerDacpacDatabaseModelFactory.cs#L56 ?

ErikEJ:
>> but it's not finding any tables. 

@bricelam This is fixed in latest master! https://github.com/ErikEJ/EFCorePowerTools/commit/5ac0582bd78ca72d45c0c77121eae9c58af7f4e0

bricelam:
Works like a charm! I wonder if there's anything we can do about the connection string:
``` csharp
protected override void OnConfiguring(DbContextOptionsBuilder options)
    => options.UseSqlServer("..\\ChinookDatabase\\bin\\Debug\\ChinookDatabase.dacpac");
```

ErikEJ:
Not sure what you mean??

ErikEJ:
Do you mean something like

"DacpacPath=..\folder\database.dacpac" ?

bricelam:
That's the code I get when I execute the command in the description. We'd need a way to let IDatabaseModelFactory implementations alter the scaffolded connection string (probably via an annotation on DatabaseModel) to make it better. Outside the scope of this PR. 🙂

bricelam:
Ideally, I want it to generate something like `Data Source=(local);Initial Catalog=ChinookDatabase` even tough I passed a dacpac to Scaffold-DbContext

bricelam:
Filed https://github.com/dotnet/efcore/issues/19799

ErikEJ:
Just published a package for EF Core 5 preview 6

Issue #274, opened 2019-10-08
Support composite database projects


Herdo:
This issue is a rough idea for reverse engineering a database context from a composite database project.

A composite database project is the aggregate of multiple, smaller and reusable DACPACs. For example, a shared DACPAC contains a table called `User`, which is used accross multiple applications that access the same database. Instead of defining the table `User` in every application's database project, the table is provided in the DACPAC. This DACPAC gets referenced by the application database projects.

When reverse engineering the composite application database project, I'd like to include the `User` table from the shared DACPAC in the same data context.

---
The only way to do this right now is to deploy the composite database project with all dependencies to a database, and reverse engineer from there.

With this issue, I'd like to suggest to "somehow" skip this deployment step, and directly create a composite database context from the composite database project.

ErikEJ:
Can you provide a sample VS solution?

Herdo:
> Can you provide a sample VS solution?

Sure, will do soon.

Herdo:
Here's an example solution:
[CompositeDatabase.zip](https://github.com/ErikEJ/EFCorePowerTools/files/3718258/CompositeDatabase.zip)

Trying to reverse engineer the composite database project fails for me.

Herdo:
Note to the ZIP: I didn't expect it to fail in the initial report. So I'd still call this an enhancement than a bug 😉

ErikEJ:
Hmm.. Looks like you can:
- Find any external references in the "mother" dacpac:
```
		<CustomData Category="Reference" Type="SqlSchema">
			<Metadata Name="FileName" Value="C:\USERS\ERIK\DOWNLOADS\COMPOSITEDATABASE\COMPOSITEDATABASE\SHAREDDATABASEOBJECTS\BIN\DEBUG\SHAREDDATABASEOBJECTS.DACPAC" />
```
https://github.com/GoEddie/Dacpac-References 

If you find any then you can load the reference dacpac, script it, and load it into the "mother" TSqlModel, but it seems quite brittle...

https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.dac.dacservices.script?view=sql-dacfx-140.3881.1

https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.dac.model.tsqlmodel.addobjects?view=sql-dacfx-140.3881.1#Microsoft_SqlServer_Dac_Model_TSqlModel_AddObjects_System_String_

Herdo:
[...]

PR #369, opened 2020-04-16
Fix for case 2 in #274

Herdo:
parser.GetCustomData() throws FileNotFoundException for missing DACPACs.

The FileNotFoundException is OK if the reference of the root DACPAC is not found.
It is however not OK if the missing DACPAC is only referenced by another DACPAC than the root DACPAC, but not by the root DACPAC directly.
Therefore, skip missing DACPACs if they're not directly referenced by the root DACPAC.

### DIFF
--- a/src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
+++ b/src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
@@ -39,8 +39,8 @@ namespace ReverseEngineer20.DacpacConsolidate
             var parser = new HeaderParser(dacpacPath);
 
             var references = parser.GetCustomData()
-                                   .Where(p => p.Category == "Reference" && p.Type == "SqlSchema")
-                                   .ToList();
+                .Where(p => p.Category == "Reference" && p.Type == "SqlSchema")
+                .ToList();
 
             if (references.Count == 0)
             {
--- a/src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
+++ b/src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
@@ -10,7 +10,7 @@ namespace ReverseEngineer20.DacpacConsolidate
     {
         public string Consolidate(string dacpacPath)
         {
-            var fileNames = GetAllReferences(dacpacPath);
+            var fileNames = GetAllReferences(dacpacPath, true);
 
             if (fileNames.Count == 0)
             {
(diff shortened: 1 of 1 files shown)


src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
```diff
@@ -26,13 +26,21 @@ public string Consolidate(string dacpacPath)
...
+                // For that case, DACPACs not referenced directly by the root DACPAC are optional.
+                return Enumerable.Empty<string>().ToList();
+            }
+
             var parser = new HeaderParser(dacpacPath);
 
             var references = parser.GetCustomData()
-                .Where(p => p.Category == "Reference" && p.Type == "SqlSchema")
-                .ToList();
+                                   .Where(p => p.Category == "Reference" && p.Type == "SqlSchema")
```
ErikEJ, line 42: Please fix indentation

ErikEJ:
Looks good, assume you have verified? Just small formatting issue

src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
Herdo: Thank you ReSharper. Will fix.

Herdo:
> Looks good, assume you have verified? Just small formatting issue

Currently on it - fails to build on my work machine, though.
I'll give you a GO when it works.

ErikEJ:
Otherwise just fix the formatting, and I will run some testing after merge

Herdo:
Okay, go ahead. **I'll try verify the merge with the new daily build.**
Local version is missing some manifest stuff I'm unable to find anywhere.

ErikEJ:
Could you provide details about the build issue, it is important for me that contributors can build without friction?

(I will also update the efreveng.exe.zip file after merge, so hold on a little while with the daily build)

ErikEJ:
Daily build 2.4.30 is now available

Herdo:
@ErikEJ Build error on work machine was: `The target “GenerateFileManifest” does not exist in the project.`

I tried to verify the changes with the new daily build. However, the next error pops up now.
I'll need to debug this to solve that. Could you please have a look after the error above?

ErikEJ:
@Herdo  Could you share your stack of files with me privately?
What is "the next error" ?
Re the error:
Do yo have the VS extensibility workload installed?
Which VS version do you use?

ErikEJ:
@Herdo Ping?

Herdo:
I got it to compile - currently working on the fix(es).

Visual Studio Version is 16.5.4. Visual Studio extension development workload is installed.
Building as x86 in Debug mode, with EFCorePowerTools as startup project. 

Now it’s working – error is gone. Only difference was restarting Visual Studio a few times.
No changes were made otherwise. Let’s just dismiss this as random Visual Studio error. 

I'll open two new pull requests for the fixes later.

Herdo:
@ErikEJ Pong

Herdo:
@ErikEJ Turns out "the next error" was a very weird state in several DACPACs.
It get's quiet complicated, and I conclude that there's no error in the current state of the EFCorePowerTools, instead our sqlprojects at work were messed up.

I'll make a new pull request with some details.

PR #371, opened 2020-04-16
Stability improvements and developer help

Herdo:
Follow-up to #274 and #369 

Stability Improvements:
- MAJOR: If a referenced DACPAC points to other DACPACs, and those DACPACs don't exist in the build output of our main project, it was included in the additional files. With this change, missing third-level DACPACs won't be included in the final reference list.
- MINOR: While building, the last write time might exceed two seconds, but the file was still built correctly. Validate against the build start time instead.
- MINOR: Accessing project.FullName might throw a NotImplementedException (from COM). In that case, the best match we can find is using the UniqueName.

Improved developer help:
- Added additional exception information for missing required references. This issues might be caused by an inconsistent usage of SuppressMissingDependenciesErrors in the database projects. The improved error message will point out that this might be the reason for the exception.

### DIFF
--- a/src/GUI/EFCorePowerTools/Extensions/EnvDTEExtensions.cs
+++ b/src/GUI/EFCorePowerTools/Extensions/EnvDTEExtensions.cs
@@ -65,12 +65,14 @@ namespace EFCorePowerTools.Extensions
                 File.Delete(file);
             }
 
+            var buildStartTime = DateTime.Now;
             if (!project.TryBuild()) return null;
 
             files = DirSearch(searchPath, "*.dacpac");
             foreach (var file in files)
             {
-                if (File.GetLastWriteTime(file) > DateTime.Now.AddSeconds(-2))
+                var lastWriteTime = File.GetLastWriteTime(file);
+                if (lastWriteTime > buildStartTime)
                 {
                     return file;
                 }
@@ -84,9 +86,21 @@ namespace EFCorePowerTools.Extensions
             var projects = Projects(dte);
             foreach (var project in projects)
             {
-                if (project.FullName == projectItemPath)
+                // Accessing project.FullName might throw a NotImplementedException (from COM).
+                // In that case, the best match we can find is using the UniqueName.
+                try
                 {
-                    return project;
+                    if (project.FullName == projectItemPath)
+                    {
+                        return project;
+                    }
+                }
+                catch (NotImplementedException)
+                {
+                    if (projectItemPath.EndsWith(project.UniqueName))
+                    {
+                        return project;
+                    }
                 }
             }
             return null;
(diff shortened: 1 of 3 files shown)


src/GUI/EFCorePowerTools/Extensions/EnvDTEExtensions.cs
```diff
@@ -84,9 +86,21 @@ private static Project GetProject(DTE dte, string projectItemPath)
             var projects = Projects(dte);
             foreach (var project in projects)
             {
-                if (project.FullName == projectItemPath)
```
ErikEJ, line 87: Is this something you have faced?

src/GUI/EFCorePowerTools/Extensions/EnvDTEExtensions.cs
```diff
@@ -65,12 +65,14 @@ public static string BuildSqlProj(this DTE dte, string sqlprojPath)
                 File.Delete(file);
             }
 
+            var buildStartTime = DateTime.Now;
```
ErikEJ, line 68: Good fix!

src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
```diff
@@ -54,6 +54,14 @@ private List<string> GetAllReferences(string dacpacPath, bool isRootDacpac)
                 .Select(i => Path.GetFullPath(Path.Combine(Path.GetDirectoryName(dacpacPath), i.Value)))
                 .ToList();
 
+            if (!isRootDacpac)
```
ErikEJ, line 57: Do we not already handle this in DcapacMerger line 39?

src/GUI/EFCorePowerTools/Extensions/EnvDTEExtensions.cs
```diff
@@ -65,12 +65,14 @@ public static string BuildSqlProj(this DTE dte, string sqlprojPath)
...
             if (!project.TryBuild()) return null;
 
             files = DirSearch(searchPath, "*.dacpac");
             foreach (var file in files)
             {
-                if (File.GetLastWriteTime(file) > DateTime.Now.AddSeconds(-2))
```
ErikEJ, line 73: Nice fix!

src/GUI/EFCorePowerTools/Extensions/EnvDTEExtensions.cs
ErikEJ, line 87: Is this something you have faced in real life?

Herdo: Indeed, while debugging earlier today.
It would randomly fail and threw NotImplementedExceptions when accessing the `FullName` property, both from the compiled code and the IntelliSense.
Therefore the fallback using the unique name.

src/GUI/ReverseEngineer20/DacpacConsolidate/DacpacConsolidator.cs
Herdo: Indeed, I didn't see that one.
Could either remove this, or keep it to "fail fast" and remove missing paths earlier.
Sadly I forgot the exact reason I put that there, or rather why the line in DacpacMerger didn't solve the issue I had.

PR #471, opened 2020-08-20
Add schema picker to filter schemas in table selection

ProH4Ck:
I added a feature to select custom schemas before table selection.
Actually only the user schema is loaded (at least with Oracle).


ErikEJ:
Thanks, maybe you should have proposed and discussed this first?

Not all databases use schemas.

and seems like an unnecessary step - maybe add a filtering option to the table selection dialog instead?

ProH4Ck:
Hi, @ErikEJ ,
When you are on the table picker dialog it's too late because tables are already loaded and if you don't pass `Schemas` property on `IDatabaseModelFactory` it loads only user tables. 
For databases that don't support schemas the user will never check the "Filter schemas" checkbox and everything works as now...

ErikEJ:
It sounds like you are working around a bug with the Oracle provider - selecting schemas works fine for SQL Server with the current design !?

ErikEJ:
Please share screenshots of the modified flow, thanks!

ProH4Ck:
So for SQL it loads all tables of all schemas the user has access to? 
With Oracle if you don't pass anything (default behavior) it loads only user schema (in Oracle a user always has a schema called with his name).

ChooseDb
ChooseSchemas
ChooseTables

ErikEJ:
Yes, that is the SQL server behaviour. 

Looks good, make sure Filter Schemas is not selected by default on the initial dialog.

ProH4Ck:
Yes. It's unchecked by default and setting is persisted (along with selected schemas) on Json configuration file.

src/GUI/EFCorePowerTools/EFCorePowerToolsPackage.cs
```diff
@@ -245,15 +245,18 @@ private IServiceProvider CreateServiceProvider()
                     .AddTransient<IPickServerDatabaseDialog, PickServerDatabaseDialog>()
                     .AddTransient<IPickTablesDialog, PickTablesDialog>()
                     .AddTransient<IModelingOptionsDialog, EfCoreModelDialog>()
-                    .AddTransient<IMigrationOptionsDialog, EfCoreMigrationsDialog>();
+                    .AddTransient<IMigrationOptionsDialog, EfCoreMigrationsDialog>()
+                    .AddTransient<IPickSchemasDialog, PickSchemasDialog>()
+                    .AddSingleton<Func<IPickSchemasDialog>>(sp => sp.GetService<IPickSchemasDialog>);
```
ErikEJ, line 250: Why is this required for the new dialog?

ProH4Ck: To avoid ServiceLocator (anti)pattern.
If you inject directly `IPickSchemasDialog` into `IPickServerDatabaseViewModel` it works only once and you got an error if you close and re-open Schemas dialog (without closing PickServerDatabaseDialog). This happens because in `IPickServerDatabaseViewModel` only a single istance of `IPickSchemasDialog` is created; when this instance is closed invoking `CloseRequested` you cannot use it anymore. To create a new istance every time dialog is requested (like you already do with other dialogs) a factory function is the only solution.

ErikEJ:
Just to be clear - you figured out how to build the .NET Core external tool (efreveng.exe) ?

ErikEJ:
Did some UI testing, the entry of schemas seem quite quirky - but works though.

I was allowed to enter an empty schema name, causing a NRE here:

            if (schemas != null)
                arguments += $" \"{string.Join(",", schemas.Select(s => **s.Name**.Replace("\"", "\\\"")))}\"";

src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
```diff
@@ -72,8 +75,10 @@ public async void ReverseEngineerCodeFirst(Project project)
                     return;
 
                 var useEFCore5 = pickDataSourceResult.Payload.IncludeViews;
+                var filterSchemas = pickDataSourceResult.Payload.FilterSchemas;
+                var schemas = filterSchemas ? pickDataSourceResult.Payload.Schemas : null;
 
-                _package.Dte2.StatusBar.Text = "Loading schema information...";
+                _package.Dte2.StatusBar.Text = "Loading schema(s) information...";
```
ErikEJ, line 81: Please revert this line

ErikEJ: Actually change to: "Loading table list..."

ProH4Ck: Fixed

ProH4Ck:
> Did some UI testing, the entry of schemas seem quite quirky - but works though.
> 
> I was allowed to enter an empty schema name, causing a NRE here:
> 
> ```
>         if (schemas != null)
>             arguments += $" \"{string.Join(",", schemas.Select(s => **s.Name**.Replace("\"", "\\\"")))}\"";
> ```

Fixed & test added

PR #484, opened 2020-08-26
AsNoTracking() for procedures call

### LINKED ISSUE #483: Consider adding AsNoTracking() to StoredProcedures call
https://github.com/ErikEJ/EFCorePowerTools/blob/26f1cbc99c457f82b38ce1f91d6e201e98562f7e/src/GUI/RevEng.Core/DbContextExtensions#L14

consider to add an `AsNoTracking()` before the `ToArrayAsync()` as the [ef core docs states that](https://docs.microsoft.com/en-us/ef/core/querying/raw-sql#change-tracking):

>Queries that use the FromSqlRaw or FromSqlInterpolated methods follow the exact same change tracking rules as any other LINQ query in EF Core.

This can improve query with a lot of data, avoid to consume too much memory and cpu to manage the tracking of all the records.

micdenny:
fixes #483

### DIFF
--- a/src/GUI/RevEng.Core/DbContextExtensions
+++ b/src/GUI/RevEng.Core/DbContextExtensions
@@ -11,7 +11,7 @@ public static class DbContextExtensions
     {
         using (var db2 = new ContextForQueryType<T>(db.Database.GetDbConnection(), db.Database.CurrentTransaction))
         {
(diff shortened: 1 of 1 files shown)


micdenny:
Not needed because of this:

>Keyless entity types are never tracked. Wherever this article mentions entity types, it refers to entity types which have a key defined.

https://docs.microsoft.com/en-us/ef/core/querying/tracking

Issue #160, opened 2019-01-09
bit not null created as bool?


mike-upshon:
When a table has column with type bit not null EFCore PowerTools generates as bool? instead of bool

### Steps to reproduce
Create table with bit not null column and reverse engineer

### Further technical details
EF Core Power Tools version: 2.2.2.0

Database engine: SqlServer

Visual Studio version: Visual Studio Ent 2017 15.9.5

ErikEJ:
Hmmm. That should have been fixed in 2.2 - can you please share a CREATE TABLE script, thanks

mike-upshon:
Sorry for the delay have been travelling.
`
USE [Test]
GO

/****** Object:  Table [dbo].[Test]    Script Date: 11-Jan-19 11:37:32 AM ******/
SET ANSI_NULLS ON
GO

SET QUOTED_IDENTIFIER ON
GO

CREATE TABLE [dbo].[Test](
	[ID] [int] NOT NULL,
	[TestBool] [bit] NOT NULL,
 CONSTRAINT [PK_Test] PRIMARY KEY CLUSTERED 
(
	[ID] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
) ON [PRIMARY]
GO

ALTER TABLE [dbo].[Test] ADD  CONSTRAINT [DF_Test_TestBool]  DEFAULT ((1)) FOR [TestBool]
GO


`

Test.cs
`
using System;
using System.Collections.Generic;
using System.ComponentModel.DataAnnotations;
using System.ComponentModel.DataAnnotations.Schema;

namespace EFBitTest.Model
{
    [Table("Test")]
    public partial class Test
    {
        [Column("ID")]
        public int ID { get; set; }
        [Required]
        public bool? TestBool { get; set; }
    }
}`


efpt.config.json
`
{
  "ContextClassName": "TestContext",
  "DefaultDacpacSchema": null,
  "IdReplace": true,
  "IncludeConnectionString": true,
  "OutputPath": "Model",
  "ProjectRootNamespace": "EFBitTest",
  "SelectedToBeGenerated": 0,
  "Tables": [
    {
      "HasPrimaryKey": true,
      "Name": "dbo.Test"
    }
  ],
  "UseDatabaseNames": false,
[...]

PR #516, opened 2020-09-26
New option to tell the reverse engineering process to create non nullable boolean properties only

StefanFalz:
Added a new option GenerateNonNullableBoolsOnly (located in the AdvancedModelingOptionsDialog) to tell the reverse engineering process to create non nullable boolean properties only, no matter what kind of default value is specified for the bit column in the database.

If this option is set, each not nullable bit column with a sql default value other than 0 is mapped as:

```csharp
public bool IsActive { get; set; }
```

instead of:

```csharp
[Required]
public bool? IsActive { get; set; }
```

If there is no default value or a value of 0 specified, the scaffolder already generates public bool IsActive { get; set; }.

I think, there could a bunch of people which will like this option.

### DIFF
--- a/src/GUI/EFCorePowerTools.Shared/Models/ModelingOptionsModel.cs
+++ b/src/GUI/EFCorePowerTools.Shared/Models/ModelingOptionsModel.cs
@@ -30,6 +30,7 @@
         private bool _mapSpatialTypes;
         private bool _mapNodaTimeTypes;
         private bool _useEf6Pluralizer;
+        private bool _generateNonNullableBoolsOnly;
 
         public bool UseDataAnnotations
         {
@@ -258,6 +259,17 @@
             {
                 if (value == _useEf6Pluralizer) return;
                 _useEf6Pluralizer = value;
+                OnPropertyChanged();
+            }
+        }
+
+        public bool GenerateNonNullableBoolsOnly
+        {
+            get => _generateNonNullableBoolsOnly;
+            set
+            {
+                if (value == _generateNonNullableBoolsOnly) return;
+                _generateNonNullableBoolsOnly = value;
                 OnPropertyChanged();
             }
         }
--- a/src/GUI/EFCorePowerTools/Dialogs/AdvancedModelingOptionsDialog.xaml
+++ b/src/GUI/EFCorePowerTools/Dialogs/AdvancedModelingOptionsDialog.xaml
@@ -68,6 +68,11 @@
                       Content="Use EF6 pluralizer"
                       IsChecked="{Binding Model.UseEf6Pluralizer}" 
                       Style="{StaticResource MarginCheckBlockStyle}"/>
+
+            <CheckBox TabIndex="5"
+                      Content="Always generate non nullable bools"
+                      IsChecked="{Binding Model.GenerateNonNullableBoolsOnly}" 
+                      Style="{StaticResource MarginCheckBlockStyle}"/>
         </StackPanel>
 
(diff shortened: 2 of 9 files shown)


ErikEJ:
What if you are not using DataAnnotations ?

ErikEJ:
What is it with (SF) in the vsct file,??

StefanFalz:
> 
> 
> What is it with (SF) in the vsct file,??

Sorry, my fault. I thought I'd removed that. It's done and commited now.

>
>
> What if you are not using DataAnnotations ?

In this case I don't see any way to retrieve the information whether the bit column in the database is nullable or not in the ReverseEngineerRunner class.

At the first look I thought the neccessary changes in the project could be made in the CommentCSharpEntityTypeGenerator class but the methods in this class are never executed? The GenerateProperties and GenerateRequiredAttribute would have been great places for the changes.

Ok, one could execute some SQL scripts against the database to check whether the column is nullable or not but I don't think this is the right way.

Do you have a suggestion how this could be solved?

---

It's ok if you decline the pull request. I'm new to Git and perhaps I accidentally didn't follow some rules. But I promise I am working on it. :)

ErikEJ:
Sorry, but I am not going to accept this - see the explanation here: https://github.com/dotnet/efcore/issues/10840#issuecomment-362047565 - so the correct resolution if you must have non-nullable properties is to drop the constraint.

Maybe it is a good idea to discsuss a PR before spending time on the actual code?

StefanFalz:
>
> 
> Maybe it is a good idea to discsuss a PR before spending time on the actual code?

Your are right. But in my case the spended time wasn't wasted. I can use the adapted version in my local environment for now and that's ok for me. To avoid problems with future versions of your tools, I'll outsource the changes into a separate project.

I understand the explanation from @bricelam but our devs see it the same way I do.

Thank you for your time and again for your excellent extension.

ErikEJ:
@StefanFalz Fine, but I do not understand how you will then need the database default.

If you use a non nullable property, EF Core will always set the value to 0 unless you explicitly set it to 1,  and if you do that, the database default makes no sense !?

rationalmonkey:
ErikEJ, why are you viewing it like Entity Framework is what those database constraints are there for in the first place.  Our project just tried to convert our primary database to EF Core (instead of EF6) and this single issue is the final issue that prevented it.  Our 2 newer databases already use EF Core and we were trying to finally end-of-life EF6.  We simply cannot make these columns nullable in C# (that's not our business layer's intention) and we cannot remove the default constraints (those are used by our seed data pipeline, our stored procs and other code that interacts with the database that is NOT using EF Core to do so).  So the inability to have EFCore generate the same C# classes and behaviors that previous EF tools supported (for instance EF6) is a problem for us.

You rejecting the OPTION of letting clients generate the code they need is very unfortunate.  Your opinion that they shouldn't want the feature, does not hold weight in whether or not the feature is useful.  If you simply google for this error message you will see no less than 5 places where other threads by different authors are seeking this feature.

ErikEJ:
@rationalmonkey As stated above, this PR is not a valid fix, and as a maintainer I am allowed to refuse invalid fixes.

Let me think about the correct fix, based on this requirement:
**an option for all non-nullable bools to not have any defaults defined** 

@rationalmonkey I assume that would solve your issue?

In the meantime, you can try to scaffold your database with a user that does not have VIEW DEFINITION rights.

ErikEJ:
@rationalmonkey I have re-opened #160

PR #578, opened 2020-11-13
Add password prompt back that was lost during a refactor

### LINKED ISSUE #575: Unable to connect to database with SQL authentication that has special characters in the password
I have am dealing with a database that only support SQL authentication and the password I was given has the special characters '$' and ';'. When using scaffold-dbcontext the only workaround that worked for me was to wrap the data-source in single-quotes and the password in double quotes:

Scaffold-DbContext 'Data Source=whatever.net,1433;Initial Catalog=whatEverDb;User Id=MyUserId;Password="dumbPas$word;";' ...

 However I want to use EFPT to do some column renaming, but I cannot connect through the EFPT GUI as the password is presumably not being sanitized in a way that allows special characters.

### Steps to reproduce
1. Create a test database in MS-SQL-Server with a SQL authenticated password like 'dumbPas$word;'. 
2. Try connecting through Reverse-Engineering Menu. Should be unable to connect. 

### Further technical details
EF Core Power Tools version: 2.5.277

Database engine: SQL Server 

Visual Studio version: Visual Studio 2019 16.7.7

ErikEJ:
fixes #575

### DIFF
--- a/src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
+++ b/src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
@@ -14,6 +14,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using ReverseEngineer20.ReverseEngineer;
+using Microsoft.VisualStudio.Data.Services;
 
 namespace EFCorePowerTools.Handlers
 {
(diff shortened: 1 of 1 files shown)


ErikEJ:
See https://github.com/ErikEJ/EFCorePowerTools/commit/ae3b449fbda79a3ec1f54191815b4a099b81c07b

PR #582, opened 2020-11-15
Adds ability to exclude columns from mapping during reveng

sm15455:
The modifications are quite simple. I have added the ExcludedColumns in the TableInformationModel. 
The awkward part is in the ReverseEngineerHandler. I had to replace predefined tables with preselected ones when the names match. This is necessary because the UI serializes from predefined tables and that would overwrite any change made by the user to the efpt.config.json file. If you have a better idea I'm all ears :)

### DIFF
--- a/src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
+++ b/src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
@@ -148,12 +148,13 @@ namespace EFCorePowerTools.Handlers
                     }
                 }
 
-                for (var i = 0; i < predefinedTables.Count; i++)
-                {
-                    var preselectedTable = preselectedTables.FirstOrDefault(c => c.Name.Equals(predefinedTables[i].Name, StringComparison.OrdinalIgnoreCase));
-                    if (preselectedTable != null)
-                        predefinedTables[i] = preselectedTable;
-                }
+                //TODO: check if there's a better way to not erase excluded columns.
+                //for (var i = 0; i < predefinedTables.Count; i++)
+                //{
+                //    var preselectedTable = preselectedTables.FirstOrDefault(c => c.Name.Equals(predefinedTables[i].Name, StringComparison.OrdinalIgnoreCase));
+                //    if (preselectedTable != null)
+                //        predefinedTables[i] = preselectedTable;
+                //}
 
                 var ptd = _package.GetView<IPickTablesDialog>()
                                   .AddTables(predefinedTables)
--- a/src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
+++ b/src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
@@ -148,10 +148,17 @@ namespace EFCorePowerTools.Handlers
                     }
                 }
 
+                for (var i = 0; i < predefinedTables.Count; i++)
+                {
+                    var preselectedTable = preselectedTables.FirstOrDefault(c => c.Name.Equals(predefinedTables[i].Name, StringComparison.OrdinalIgnoreCase));
+                    if (preselectedTable != null)
+                        predefinedTables[i] = preselectedTable;
+                }
+
                 var ptd = _package.GetView<IPickTablesDialog>()
                                   .AddTables(predefinedTables)
                                   .PreselectTables(preselectedTables);
-                
+
                 var customNameOptions = CustomNameOptionsExtensions.TryRead(renamingPath);
 
                 var pickTablesResult = ptd.ShowAndAwaitUserResponse(true);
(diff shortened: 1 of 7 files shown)


src/GUI/RevEng.Core/VisitorRelationScaffoldingModelFactory.cs
```diff
@@ -0,0 +1,55 @@
...
+			else if (_databaseType == DatabaseType.SQLServer)
+			{
+				fullTableName = $"[{table.Schema}].[{table.Name}]";
+			}
+			else
+			{
+				fullTableName = $"{table.Schema}.{table.Name}";
+			}
+			var tableDefinition = _tables.FirstOrDefault(c => c.Name.Equals(fullTableName, StringComparison.OrdinalIgnoreCase));
+			if (tableDefinition != null && tableDefinition?.ExcludedColumns != null) {
```
ErikEJ, line 44: Could this just be: if (tableDefinition?.ExcludedColumns != null) {

src/GUI/RevEng.Core/VisitorRelationScaffoldingModelFactory.cs
```diff
@@ -0,0 +1,55 @@
...
+				foreach (var column in tableDefinition?.ExcludedColumns) {
```
ErikEJ, line 45: Please add new line before "{"

src/GUI/RevEng.Shared/TableInformationModel.cs
```diff
@@ -19,6 +20,7 @@ public class TableInformationModel : INotifyPropertyChanged
         private bool _hasPrimaryKey;
         private bool _showKeylessWarning;
         private ObjectType _objectType;
+        private IEnumerable<string> _excludedColumns = new List<string>();
```
ErikEJ, line 23: Will this add empty ExcludedColumns arrays to all tables when serialized?

src/GUI/RevEng.Shared/TableInformationModel.cs
```diff
@@ -89,6 +91,21 @@ public ObjectType ObjectType
             }
         }
 
+        /// <summary>
+        /// Gets or sets the object type.
```
ErikEJ, line 95: Fix comment

src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
```diff
@@ -148,10 +148,17 @@ public async System.Threading.Tasks.Task ReverseEngineerCodeFirstAsync(Project p
                     }
                 }
 
+                for (var i = 0; i < predefinedTables.Count; i++)
```
ErikEJ, line 151: Suggest you comment this code out with a TODO, and I will investigate once this is merged,

ErikEJ (changes requested):

ErikEJ:
Pls remove the two .zip files, I will update after merge

src/GUI/RevEng.Shared/TableInformationModel.cs
sm15455: yes

src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
sm15455: If I comment this code, the reveng process would overwrite modifications to efpt.config.json file. Is it ok?

sm15455:
Regarding the .zip files, should I revert the commits or should I commit original version?

ErikEJ:
Revert pls

src/GUI/RevEng.Shared/TableInformationModel.cs
ErikEJ: Could we avoid that?

src/GUI/EFCorePowerTools/Handlers/ReverseEngineerHandler.cs
ErikEJ: Yes, I will do some testing when this has been merged

src/GUI/RevEng.Shared/TableInformationModel.cs
sm15455: I removed initialization and added EmitDefaultValues = false to DataMember. It still serializes the null value.
```
{
  ExcludedColumns: null
   ...
}
```

It shouldn't be a problem, but I'll keep investigating.

ErikEJ: Never mind...

ErikEJ (approved):

ErikEJ:
LGTM - do you have anything to add?

sm15455:
Nothing to add. I'll work on the UI side on another branch if you think it's ok, but first we would need to discuss about this.

ErikEJ:
Sounds great, let me just think about the UI design first.

sm15455:
Ok, once you merge the code into master I'll open a new issue so we can discuss about it.

ErikEJ:
Thanks for a great contribution! (so far 😄 )

Issue #588, opened 2020-11-17
Add UI to exclude columns from reverse engineering


ErikEJ:
- [x] Create new class to exchange the table list (stop using TableInformation).
- [x] Add columns to this class.
- [x] Add columns to the output from TableList command if EFPT CLI tool.
- [x] Extend TableInformation with a new, non-serialized `List<ColumnInformation>`, where ColumnInformation has: string Name, bool Excluded
- [x] Construct the ExcludeColumns property when loading options, and set when saving options.
- [x] Bind ColumnInfomation to UI - no idea how...

@sm15455 FYI

sm15455:
These steps looks ok to me. Regarding binding to UI, a treeview with checkboxes should work. Do you know if treeview virtualizes rendering? I'm afraid that for database with lots of tables and columns, rendering could slow slow down interface. In that case we could load columns on demand.

ErikEJ:
I do not know much about WPF, but lazy loading the columns is maybe something we should just do anyway.

sm15455:
Deal. Do you want me to work on the missing points?

sm15455:
TableInformationViewModel has TableInformationModel property and a Selected property. If I add a Columns property (list of string [name], bool [selected]) to TableInformationModel we will have Selected property in the model for columns and Selected property  in the view model for tables. This looks weird to me. 

I still haven't fully analyzed the code, but I was thinking to:
- Remove TableInformationModel in TableInformationViewModel.
- Splatter TableInformationModel properties in TableInformationViewModel.
- Create a ColumnInformationViewModel class with Name and Selected Property.
- Add a Columns (List<ColumnInformationViewModel>) property to TableInformationViewModel.
- Change the AddTable and SelectTables methods in TableInformationViewModel to accept a TableModel parameter and populate the properties of TableInformationViewModel.
- Change the GetResult methods in TableInformationViewModel to return a list of TableModel

This would make thing less coupled between models and view models but would require lots of changes in code. What do you think?

ErikEJ:
I think it would be a healthy refactor - maybe you could do that in a separate PR before making any actual UI changes?

sm15455:
If I make a PR with these changes only, UI would be in inconsistent state. I should at least make the minimal changes to align it with the TableInformationViewModel changes.

ErikEJ:
Got it!

sm15455:
[...]

PR #589, opened 2020-11-17
Return columns from table list

ErikEJ:
add missing MySQL table list support

relates to #588

@sm15455 - can you have a look?


sm15455:
Looks good to me

PR #607, opened 2020-11-23
Add more images to table list and move them in front

ErikEJ:


ErikEJ:
@sm15455 FYI

sm15455:
Looks good. Maybe TableModel should be renamed to ObjectModel.

ErikEJ:
image

ErikEJ:
If we redesign more, we can rename classes and refactor HasPrimaryKey

sm15455:
Ok, I'll try to do that when I'll implement object groups in treeview.

ErikEJ:
image

sm15455:
The good old EF6 experience... With grouping we'll be there :D

ErikEJ:
The distinction between views and tables is more blurry with EF Core, though.

ErikEJ:
I also think the "schema" (dbo) level is unnecessary.

sm15455:
You are right. I would avoid nesting per schema. We can always implement it later if users ask for it. You are also right about viees and tables in EF Core, but this view organizes database objects and there the classification is clear.

PR #637, opened 2020-12-07
Reverse Engineering - Always return TEntity[] from SP mapper

### LINKED ISSUE #636: [Proposal] Reverse Engineering - Always return TEntity[] from SP mapper
At this point if an output column list could not be retrieved (SP uses temp table/dynamic SQL, etc), SP mapper returns Task<int>. SP mapper body calls ExecuteSqlRawAsync.
Other SP mappers return Task<procedurenameRESULT[]>. Since procedurenameRESULT class is partial, user can extend this class with custom properties in order to obtain data from SP with temp tables/dynamic SQL.

My proposal is to use SqlQueryAsync&lt;procedurenameRESULT&gt; despite the errors while retrieving data structure.

### Further technical details
EF Core Power Tools version: 2.5.358

Database engine: MS SQL 

Visual Studio version: Visual Studio 2019 Professional v 16.8.2.

tiomny:
Resolves #636.

### DIFF
--- a/src/GUI/RevEng.Core/Procedures/SqlServerStoredProcedureScaffolder.cs
+++ b/src/GUI/RevEng.Core/Procedures/SqlServerStoredProcedureScaffolder.cs
@@ -224,15 +224,8 @@ namespace RevEng.Core.Procedures
                         }
                     }
                     _sb.AppendLine("};");
-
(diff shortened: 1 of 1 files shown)


ErikEJ:
LGTM

PR #639, opened 2020-12-07
Fixed runtime exception

tiomny:
No mapping to a relational type can be found for the CLR type 'SqlParameter[]

### DIFF
--- a/src/GUI/RevEng.Core/DbContextExtensions
+++ b/src/GUI/RevEng.Core/DbContextExtensions
@@ -3,8 +3,8 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Storage;
 using System;
-using System.Collections.Generic;
 using System.Data.Common;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
--- a/src/GUI/RevEng.Core/DbContextExtensions
+++ b/src/GUI/RevEng.Core/DbContextExtensions
@@ -14,6 +14,8 @@ namespace #NAMESPACE#
     {
         public static async Task<T[]> SqlQueryAsync<T>(this DbContext db, string sql, object[] parameters = null, CancellationToken cancellationToken = default) where T : class
         {
+            parameters ??= new object[] { };
+
             using (var db2 = new ContextForQueryType<T>(db.Database.GetDbConnection(), db.Database.CurrentTransaction))
             {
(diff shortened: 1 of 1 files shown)


tiomny:
Hold a minute - now I get "Value cannot be null. (Parameter 'parameters')"

tiomny:
Sorry for the late full manual testing - it seems fine now. @ErikEJ could you merge it?

src/GUI/RevEng.Core/DbContextExtensions
```diff
@@ -12,8 +12,10 @@ namespace #NAMESPACE#
 {
     public static class DbContextExtensions
     {
-        public static async Task<T[]> SqlQueryAsync<T>(this DbContext db, string sql, IEnumerable<object> parameters, CancellationToken cancellationToken = default) where T : class
+        public static async Task<T[]> SqlQueryAsync<T>(this DbContext db, string sql, object[] parameters = null, CancellationToken cancellationToken = default) where T : class
         {
+            parameters ??= new object[] { };
```
ErikEJ, line 17: What is this syntax ? Will it work with C# 8 projects?

src/GUI/RevEng.Core/DbContextExtensions
```diff
@@ -12,8 +12,10 @@ namespace #NAMESPACE#
 {
     public static class DbContextExtensions
     {
-        public static async Task<T[]> SqlQueryAsync<T>(this DbContext db, string sql, IEnumerable<object> parameters, CancellationToken cancellationToken = default) where T : class
```
ErikEJ, line 15: Can we remove a using now?

ErikEJ (changes requested):

src/GUI/RevEng.Core/DbContextExtensions
tiomny: > Available in C# 8.0 and later, the null-coalescing assignment operator ??= assigns the value of its right-hand operand to its left-hand operand only if the left-hand operand evaluates to null. The ??= operator doesn't evaluate its right-hand operand if the left-hand operand evaluates to non-null.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator

ErikEJ:
Assume you have run some integration tests now 😄

ErikEJ (approved):

tiomny:
> 
> 
> Assume you have run some integration tests now 😄

This code is working in DEV environment :)

ErikEJ:
Thanks, I have just updated the tool .zip files

PR #650, opened 2020-12-10
Preserve renaming schemas without any tables

### LINKED ISSUE #645: Renaming SQL objects with schema not working anymore
Since the last update (yesterday it worked), I can't generate the models of my SQL objects with the right name anymore.
I was using the efpt.renaming.json file like this: 
```json
[
  {
    "UseSchemaName": true,
    "SchemaName": "dev"
  },
  {
    "UseSchemaName": true,
    "SchemaName": "test
  },
  {
    "UseSchemaName": true,
    "SchemaName": "prod"
  }
]
```
Before the update, it worked and generated the objects for me using the schema name as a suffix.
For example theses views: dev.myView ,test.myView ,prod.myView became myViewDev, myViewTest, myViewProd model in C#.
Now when I generate the models with the "Reverse engineer" option, my efpt.renaming.json file is empty and the models are not renamed (by default they are named myView, myView1, myView2).
My question is : how to configure the file since the update (if it is not a bug)? And how can I make it behave like before, i.e. rename all SQL object models of a schema with a suffix (stored procedure,view and table).

I also saw this comment in the wiki: ```(This file will be created for you if you use the UI based renaming feature)```.
How does it work from the interface? I don't seem to have seen the option. 
(This file will be created for you if you use the UI based renaming feature).

---

EF Core Power Tools version: 2.5.376

Database engine: SQL Server 2019

Visual Studio version: Visual Studio 2019 16.8.2

ErikEJ:
fixes #645

### DIFF
--- a/src/GUI/EFCorePowerTools/ViewModels/ObjectTreeViewModel.cs
+++ b/src/GUI/EFCorePowerTools/ViewModels/ObjectTreeViewModel.cs
@@ -125,7 +125,7 @@
                 (ObjectType.ScalarFunction, "Functions")
             };
 
-            _allSchemas = customReplacers;
+            _allSchemas = customReplacers ?? new List<Schema>();
 
             foreach (var objectType in objectTypes)
             {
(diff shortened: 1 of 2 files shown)


ErikEJ:
@sm15455 FYI!

PR #672, opened 2020-12-18
Use Task<int> and ExecuteSqlRawAsync when resultset from a stored procedure is empty and valid

### LINKED ISSUE #669: Return int type when stored procedure doing update and delete
### Issue
A stored procedure like this:
```sql
CREATE PROCEDURE [spc_User_Update] (
   @parameters
) 
AS
BEGIN
   UPDATE user
      SET name = @NAME
    WHERE ID = @ID
END
```

C# After reverse engeineer => 
```csharp
 var db = new  DBContext();
 var sp = new DBContextProcedures();
 var result = sp.spc_User_Update(@ID,@NAME).Result; // **return empty class spc_User_UpdateResult , but expected int type**
```

issue:
How to return int type when stored procedure doing update,delete command?
Thank you 👍 

### Enviroment
EF Core Power Tools version: 2.5.406.0

Database engine: SQL Server 2017

Visual Studio version: vs 2019

ErikEJ:
fixes #669


ErikEJ:
@tiomny  FYI!

PR #688, opened 2020-12-28
Add partial method to split configuration classes and make them partial

### LINKED ISSUE #686: Proposal: Add partial method for configuration in model configuration class
I have a model with several tables and I need to exclude some columns (mainly enums) that I want to manually add using partial classes. To avoid a monster context class, I also decided to split entities configuration in their own classes. 
The problem now is that I can extend generated entities using partial classes, but I need to resort to DbContext **OnModelCreatingPartial** partial method to map them correctly. I think it would be more clear if we could have an **OnConfigurePartial** method in the generated configuration classes. This way we can keep additional entity configuration in a partial close to configuration class and not in the DbContext partial.

ErikEJ:
fixes #686


ErikEJ:
@sm15455 FYI

sm15455:
Perfect!

PR #709, opened 2021-01-14
Fix schema(s) list not read from config if present

ProH4Ck:
I was using a custom version but today I updated to latest version and found this bug...


ErikEJ:
will this also persist selected schemas in the config.json file?

ProH4Ck:
Yes, persistence already works. The problem was only in initialization

PR #716, opened 2021-01-20
Enable nullable reference type option

ErikEJ:


src/GUI/RevEng.Core.50/CommentCSharpEntityTypeGenerator.cs
```diff
@@ -274,7 +285,22 @@ protected virtual void GenerateProperties([NotNull] IEntityType entityType)
...
+
+                if (_nullableReferences && !property.ClrType.IsValueType)
+                {
+                    if (property.IsColumnNullable())
+                    {
+                        nullableAnnotation = "?";
+                    }
+                    else
+                    {
+                        defaultAnnotation = $" = default!;";
```
roji, line 299: This is indeed one way to get around the property initialization problem, but the more robust/recommended way IMHO should be with a nullable backing field and a non-nullable property whose getter throws on null ([see the docs](https://docs.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types#non-nullable-properties-and-initialization)).

That's definitely more elaborate/complicated than this though...

src/GUI/RevEng.Core.50/CommentCSharpEntityTypeGenerator.cs
```diff
@@ -38,12 +39,14 @@ public class CommentCSharpEntityTypeGenerator : ICSharpEntityTypeGenerator
         /// </summary>
         public CommentCSharpEntityTypeGenerator(
             [NotNull] IAnnotationCodeGenerator annotationCodeGenerator,
-            [NotNull] ICSharpHelper cSharpHelper)
+            [NotNull] ICSharpHelper cSharpHelper,
+            bool nullableReferences)
```
roji, line 43: Note that on the EF Core side we hope to detect whether this is enabled at the project level, so ideally we wouldn't need to emit anything. But this is fine too (and much easier obviously).

roji (commented):
Looks pretty good to me! See some comments below.

src/GUI/RevEng.Core.50/CommentCSharpEntityTypeGenerator.cs
ErikEJ: Thanks, @roji I really appreciate you taking time to have a look!
For now, I will just go with this simpler approach 😄

src/GUI/RevEng.Core.50/CommentCSharpEntityTypeGenerator.cs
ErikEJ: Sure, but you would still need to pass that information to the tooling, which knows nothing about the project.

src/GUI/RevEng.Core/CommentCSharpEntityTypeGenerator.cs
```diff
@@ -261,7 +284,23 @@ private void GenerateRequiredAttribute(IProperty property)
...
                     var referencedTypeName = navigation.GetTargetType().Name;
                     var navigationType = navigation.IsCollection() ? $"ICollection<{referencedTypeName}>" : referencedTypeName;
-                    _sb.AppendLine($"public virtual {navigationType} {navigation.Name} {{ get; set; }}");
+
+                    string nullableAnnotation = string.Empty;
+                    string defaultAnnotation = string.Empty;
+
+                    if (_nullableReferences && !navigation.IsCollection())
+                    {
+                        if (navigation.ForeignKey?.IsRequired == true)
```
NickNiebling, line 293: Is this evaluation correct?
If FK (ForeignKey) is set to "Null" (nullable) in SQL, then I would assume FK.IsRequired will be "false" (not "true")

Maybe it should be "inverted"?
```
                        if (navigation.ForeignKey?.IsRequired == true)
                        {
                            defaultAnnotation = $" = default!;";
                        }
                        else
                        {
                            nullableAnnotation = "?";
                        }
```

src/GUI/RevEng.Core.50/CommentCSharpEntityTypeGenerator.cs
```diff
@@ -398,7 +424,23 @@ protected virtual void GenerateNavigationProperties([NotNull] IEntityType entity
...
                     var referencedTypeName = navigation.TargetEntityType.Name;
                     var navigationType = navigation.IsCollection ? $"ICollection<{referencedTypeName}>" : referencedTypeName;
-                    _sb.AppendLine($"public virtual {navigationType} {navigation.Name} {{ get; set; }}");
+
+                    string nullableAnnotation = string.Empty;
+                    string defaultAnnotation = string.Empty;
+
+                    if (_nullableReferences && !navigation.IsCollection)
+                    {
+                        if (navigation.ForeignKey?.IsRequired == true)
```
NickNiebling, line 433: Is this evaluation correct?
If FK (ForeignKey) is set to "Null" (nullable) in SQL, then I would assume FK.IsRequired will be "false" (not "true")

Maybe it should be "inverted"?
```
                        if (navigation.ForeignKey?.IsRequired == true)
                        {
                            defaultAnnotation = $" = default!;";
                        }
                        else
                        {
                            nullableAnnotation = "?";
                        }
```

NickNiebling (commented):
We have been trying out the experimental nullableReferences feature, and we have observed that FK (ForeignKey) handling feels wrong/inverted.

Please look at this example, where Example class has 2 FK, where one off them is "null" and the other "not null".

```
public class Example {
    public Guid ExampleId { get; set; }
    public Guid FooId { get; set; }
    public Guid? BarId { get; set; }

     public virtual Foo? Foo { get; set; }
     public virtual Bar Bar { get; set; } = default!;
}
```

I would have expected:

```
public class Example {
    public Guid ExampleId { get; set; }
    public Guid FooId { get; set; }
    public Guid? BarId { get; set; }

     public virtual Foo Foo { get; set; } = default!;
     public virtual Bar? Bar { get; set; }
}
```

ErikEJ:
Thanks for trying out this experimental feature.

Please post an issue and feel free to submit a PR.

PR #720, opened 2021-01-21
Option to ignore constructor

### LINKED ISSUE #692: Proposals: Options to remove default constructor from generated code
When an entity is generated, the default constructor is added. In some cases, especially in DDD approach, this behavior is not correct. Having the option to remove the default constructor gives us the possibility to create only the constructors we need.

I have investigated a bit and this feature could be added by creating a class that inherits from **CSharpEntityTypeGenerator** and overrides **GenerateConstructor** method. 
https://github.com/dotnet/efcore/blob/68d54dd79a798ff85f91778897087aa85de00c0e/src/EFCore.Design/Scaffolding/Internal/CSharpEntityTypeGenerator.cs#L233

ErikEJ:
fixes #692


ErikEJ:
@sm15455 FYI!

sm15455:
Looks good to me

PR #734, opened 2021-01-26
Rename Navigations (experimental)

### LINKED ISSUE #211: Reverse Engineer - Rename navigation and reference properties
Is it possible to rename the generated navigation properties?

### Steps to reproduce
Say there are two tables,

CREATE TABLE Product (
Id UNIQUEIDENTIFIER NOT NULL PRIMARY KEY, 
CategoryCode VARCHAR(20) FOREIGN KEY (CategoryCode) REFERENCES Category (Code) NULL, 
ParentId UNIQUEIDENTIFIER FOREIGN KEY (ParentId) REFERENCES Product (Id) NULL
)

CREATE TABLE Category (
Code VARCHAR(20) NOT NULL PRIMARY KEY,
Description VARCHAR(100) NOT NULL
)

Product table references both Category and Product itself. The generated class contains the following navigation properties, which I very much want to rename to something more meaningful.

`public Category CategoryCodeNavigation { get; set; }`
`public Product ParentProduct {get;set;}`
`public virtual ICollection<Product> InverseProduct {get; set;}`

I have tried to use efpt.renaming.json, but it only supports renaming columns not FK relations.

I understand I can just manually rename the properties after the code-gen, but the changes will be overwritten when the project is re-generated which we do quite often.

So is there a workaround to rename FK navigation properties?

Thanks for any tips!

### Further technical details
EF Core Power Tools version: (found in About dialog - blue questionmark icon on context menu)
2.2.12.0

Database engine: (SQlite, SQL Compact, SQL Server, Postgres)
Azure SQL

Visual Studio version: (e.g. Visual Studio 2017 15.7)
VS2019

### LINKED ISSUE #732: Navigation properties names
In our db, we don't use "Id" or "TableNameId" names for PK columns, but rather "TableNameGUID" names. After running the tool, I'm getting strange navigation property names:

Let's say we have:

* OrderLift table with
   * OrderLiftGUID
   * DriverGUID

* Driver table with
   * DriverGUID

This leads to generated OrderLift model having "DriverGu" navigation property, instead of "Driver". I tried to customize it with config.json, but that didn't work, since this is not a column in the db. I'm pretty sure this wasn't happening with EF6 or Linq2Sql.

Is there any way to customize navigation properties names?

ErikEJ:
Clean up temp files

fixes #732
fixes #211


tafs7:
@ErikEJ is this experimental feature already available in the extension installed from the VS Marketplace/Ext Manager?
I just installed the latest version (2.5.537), but don't see any UI to enable renaming navigations.
is it only currently supported via the config file, not from UI?

ErikEJ:
@tafs7 Oh yes, see the release notes with a link, as noted there is no UI for this feature (and I have a hard time figuring out how to add one, but if you have any ideas, let me hear) https://github.com/ErikEJ/EFCorePowerTools/wiki/Release-notes#25513-february-14-2021

tafs7:
@ErikEJ Thanks!  And, btw, I REALLY appreciate this tool right.  Helping out a client migrate a nearly 10-yr old EF web app that is still using EDMX, and a Database project for "migrations", and they've customized (ahem...bastardized) the EDMX T4 templates, also use a mix of stored procs, user-defined table types, functions, views, multiple db contexts - one w/ over 120 entities, and another with over 150 entities only used for working with lookup value tables (generates constants in code from table values).

I'm pretty much gonna be exploiting nearly every feature and customization in EF Core Power Tools, except nullable ref types....for now.... ;)

ErikEJ:
@tafs7 have you tried the EDMX scaffolding?

tafs7:
no - how does that work? what does it do?
I keep finding features in this tool - 1st time user here 😄 
Gonna dig thru the wiki a bit more also.

ErikEJ:
@tafs7 Similar to the .dacpac reverse engineering

ErikEJ:
(Requires latest daily!)

tafs7:
@ErikEJ just tried the EDMX scaffolding and compared with the database scaffolding.  A few differences in the output, but also, currently doesn't provide ability to scaffold stored procs and functions, so I'll just stick to the DB scaffolding for now.

WRT how you might approach building UI to support renaming navigations, I think it would be hard to accomplish this from the DB scaffolding, because you'd have to inspect foreign keys and display them as sub-nested under each table in the object selection dialog.

However, if you consider the new experimental EDMX scaffolder, it would probably be a bit easier there since you can just read the navigation properties already defined by the EDMX models without having to inspect table FKs.  You could even provide an option to either use navigations already defined in EDMX, or use default conventions for the scaffolded navigations (the behavior you get today with the database scaffolding).

That latter option would allow folks like me who have an EDMX to basically generate EF Core entities that are nearly identical to the ones generated by the EDMX T4 tool, since EPT would be reading off the same model definitions.  It would be more of a "port" from EDMX entities to Code-First entities, than a reverse-engineering process from database to code.

PR #754, opened 2021-02-06
Added UI side of database comparison feature

sm15455:
This PR adds:

- Project context menu item
- UI for picking up database and contexts
- UI that shows comparison result

It misses integration with efpt.exe which should be implemented in GetDbContextTypesAsync and GetComparisonResultAsync methods of CompareHandler class.


ErikEJ:
This is fantastic, thanks so much for your contribution!

ErikEJ:
@sm15455 I had a look at this, I think it would make sense to just limit the compare to a single DbContext for now - so maybe you can change the Options dialog to simply two dropdowns, that both must have something selected for the dialog to succeed?

sm15455:
I used a listbox to select multiple DbContexts because EFCompare library gives this chance. I can easily switch to a DropdDown. Just let me know.

ErikEJ:
Let's leave it as is for now, I also just noticed that.

PR #774, opened 2021-02-13
create procedures as extension of context

pedromcrocha:


ErikEJ:
Hmm.. this will be breaking for all current users of this feature.

ErikEJ:
Had a look, this will "pollute" intellisense for the DbContext if you have many stored procedures.

ErikEJ:
Adding this to the existing scaffolded code:

```
 public static class NorthwindContextProceduresExtensions
    {
        public static NorthwindContextProcedures Procedures(this NorthwindContext context)
        {
            return new NorthwindContextProcedures(context);
        }
    }

``` 

allows you to do this:

`var result = await db.Procedures().GetXxxAsync();`

PR #799, opened 2021-02-19
When creating stored procedures, SqlDbType.Time must map to TimeSpan

pedromcrocha:


ErikEJ:
Well spotted!
Would be great if you can build the efreveng tools - then this can go straight in!

You can use the two .cmd files to do that:

https://github.com/ErikEJ/EFCorePowerTools/tree/master/src/GUI/efreveng
and
https://github.com/ErikEJ/EFCorePowerTools/tree/master/src/GUI/efreveng50

PR #814, opened 2021-02-25
PR Add ReverseEnginner Edit/Refresh to folder

pedromcrocha:
Proposal to add ReverseEnginner Edit/Refresh to folder

### DIFF
--- a/src/GUI/EFCorePowerTools/EFCorePowerToolsPackage.cs
+++ b/src/GUI/EFCorePowerTools/EFCorePowerToolsPackage.cs
@@ -23,6 +23,7 @@ using Microsoft.VisualStudio.Shell;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Design;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


ErikEJ:
Could you explain / screenshot?

pedromcrocha:
ContextMenu in folder

detect one or more config files in folder


ContexMenuInFolder

pedromcrocha:
image

ErikEJ:
I am unsure of the value and concerned with the complexity, and have never seen a right click action on folders in other contexts with Visual Studio...

pedromcrocha:
it is just a shortcut, which I think will be appreciated by those who have many models/context.

if we assume that the folder or sub folder has a config file, we don’t need to search, just rigth click

ErikEJ:
I am not sure that this adds value, and what if there are multiple files in the folder. 

Will wait for some feedback on the brand new right click option until we expand it further.

ErikEJ:
Going to close this for now...

PR #833, opened 2021-03-04
Experiment with caching, unsure about value (hard to measure perf imp…

ErikEJ:
…rovement)

### DIFF
--- a/src/GUI/EFCorePowerTools/Handlers/ReverseEngineer/EfRevEngLauncher.cs
+++ b/src/GUI/EFCorePowerTools/Handlers/ReverseEngineer/EfRevEngLauncher.cs
@@ -51,6 +51,7 @@ namespace EFCorePowerTools.Handlers.ReverseEngineer
                 UseNoConstructor = options.UseNoConstructor,
                 UseNoNavigations = options.UseNoNavigations,
                 UseNoObjectFilter = options.UseNoObjectFilter,
+                DatabaseModelCacheTTLInSeconds = Properties.Settings.Default.ModelCacheTimeToLive,
             };
 
             var launcher = new EfRevEngLauncher(commandOptions, useEFCore5);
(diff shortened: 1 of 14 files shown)


ErikEJ:
@sm15455  Not sure this will ever go in...

sm15455:
IMO, Using a time span doesn't add value. I'll investigate more once I finish moving strings to localization files

ErikEJ:
Why not, it reduces the number of database queries by 50% during a wizard run.

sm15455:
If you use it to cache between object search and code generation then it's fine. 
I was thinking about caching between two different code generation. (I generate the code today, then I regenerate tomorrow because I need to map another table). In this last case time span is not valuable.

ErikEJ:
@sm15455 Agree, next experimental feature (with user opt-out via an option) could be schema change detection from the RDBMS. (I have already added an Options tab for setting the TTL on the cache)

sm15455:
I have made same test with different sized database. 
On my machine, the difference between a medium sized database (~100 tables, 100~ views and 600~ SP) and a small database (10 tables, 10 views and 20 SP) is not big. (less than 1 sec). The call to efreveng50.exe takes an average of 675ms for the small database and 1399 for the medium database.  (I used a laptop with an I7 processor, 32gb of RAM and 1tb SSD).
My opinions based on these numbers:
- Caching the model has not a great impact on performance. 
- Caching between object selection and code generation adds some value (using the time span pattern as you did here).
- Trying to find a smart way to detect changes on database just to persist the model cache longer (as I suggested) complicates things a lot and adds little value compared to the effort.

ErikEJ:
Thanks, I came to similar conclusion, will see if it can finalize the short term caching.

ErikEJ:
Gave up on a reliable way to support filtering, closing this given the little impact it has.

PR #836, opened 2021-03-05
Moved menu, labels, messages to resource files

sm15455:
Lots of files involved. No logic except for some string concatenation.


ErikEJ:
Wauw! This is huge...

PR #849, opened 2021-03-13
Add Simplified Chinese Localization

Varorbc:


ErikEJ:
Just a rename of existing files??

Varorbc:
Maybe I should say that only Simplified Chinese is supported

ErikEJ:
But is it not better to leave as is then? And drop this PR?

Varorbc:
Simplified Chinese and traditional Chinese are still quite different and cannot be shared

ErikEJ:
Sorry, I am an ignorant! 

Is this ready to merge?

Varorbc:
yes

PR #873, opened 2021-03-22
disabled foreign key exclusion in reverse engineer

sm15455:


src/GUI/EFCorePowerTools/ViewModels/ColumnInformationViewModel.cs
```diff
@@ -74,6 +75,18 @@ public bool IsPrimaryKey
...
 
+        public bool IsForeignKey
+        {
+            get => _isForeignKey;
+            set
+            {
+                if (Equals(value, _isForeignKey)) return;
+                _isForeignKey = value;
+                RaisePropertyChanged();
+                RaisePropertyChanged(nameof(IsEnabled));
```
ErikEJ, line 86: nameof (IsForeignKey) ? and why two x Raise ??

ErikEJ (changes requested):

src/GUI/EFCorePowerTools/ViewModels/ColumnInformationViewModel.cs
sm15455: It's correct like it is. 
The first call to RPC automatically infers the name from the property name so uses "IsForeignKey". The second call to RPC forces to reevaluate the IsEnabled property since it depends on IsForeignKey.

ErikEJ (approved):

PR #886, opened 2021-03-29
Stored procedures return list

pedromcrocha:
Proposal

Option for stored procedures return List instead of array

image


ErikEJ:
Why is that important? 
Can you just not do ToList() on the result?

pedromcrocha:
of course, but i always use list, and ef have beautiful method that retrieve  a list (ToListAsync), why not use that.

ErikEJ:
Thanks, maybe in future create an issue so we can discuss before submitting a PR (that may not get accepted) ?

pedromcrocha:
you are right, next time i will have more attention.
thanks for your patience with me.

ErikEJ:
I really appreciate your great contributions!

Issue #158, opened 2019-01-07
Reverse-Engineer with schemas into directories


mikethibault:
For reverse engineering with multiple schemas, an option to sort classes into schema-named directories would be very helpful.

ErikEJ:
Why have all your tables in a single DbContext?

Why not two projects with a DbContext per schema? (You can do that today)

ErikEJ:
Ping?

mikethibault:
My database has relationships between tables from different schemas.  I'm unaware of a way that Entity Framework can allow joins between contexts without loading everything in memory.  I think it'd be great to mirror this capability of database schemas in an EF Core context.

ErikEJ:
Not something I plan to implement, but a PR for this will be considered

Trigun27:
It will be a cool feature!

ErikEJ:
@Trigun27 PRs accepted 😀

Trigun27:
@ErikEJ Hi. I think this will be a nice feature. Maybe I can try to write the required code.
Any tips and suggestions?

ErikEJ:
Had a closer look at this, looks like an approach to try would be to implement https://github.com/dotnet/efcore/blob/main/src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs#L299 and register this override when building the service provider, if a list of columns to exclude are available, then modify the DatabaseTable model by removing the excluded columns.

we would need a way to pass the list of columns to exclude to this class. 

The list of columns to exclude could be under each Table in the tables list in efpt.config.json.

ErikEJ:
Actually, it does not look like there is any way of doing the in a generic fashion with the current implementation.

This place would be a potential place to do it, but I am not sure you have any schema information at this point: https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/RevEng.Core/ReverseEngineerRunner.cs#L116

ErikEJ:
Closing as external, must be implemented in EF Core

chassq:
Thanks for re-opening this @ErikEJ !!

ErikEJ:
I will have another look at this, could one of you outline the desired file structure?

Would it be something like:

```
.\Models
    Dbo
        Table1.cs
        Table2.cs
    Otherschema
        Table3.cs
        Table4.cs

```

chassq:
[...]

PR #981, opened 2021-05-13
Introduced schema folder separation and cleanup method improvements.

eminmesic:
I have introduced schema folder separation per issue: [https://github.com/ErikEJ/EFCorePowerTools/issues/158](https://github.com/ErikEJ/EFCorePowerTools/issues/158)

These changes include AdvancedModelingOptionsDialog and **Use schema folder separation** is added as the last option of Advance settings. You can see this in the following image. This setting is saved as UseSchemaFolders property in JSON and it is used through all changes in code.

image

There are a little bit complex changes inside **CleanUp** method because of supporting switching between schema and non-schema option and removing all directories and their files.

### DIFF
--- a/src/GUI/RevEng.Core.50/RevEng.Core.50.csproj
+++ b/src/GUI/RevEng.Core.50/RevEng.Core.50.csproj
@@ -33,7 +33,8 @@
 		<Compile Include="..\RevEng.Core\ServiceProviderBuilder.cs" Link="ServiceProviderBuilder.cs" />
 		<Compile Include="..\RevEng.Core\SharedTypeExtensions.cs" Link="SharedTypeExtensions.cs" />
 		<Compile Include="..\RevEng.Core\SqlServerHelper.cs" Link="SqlServerHelper.cs" />
-		<Compile Include="..\RevEng.Core\TableListBuilder.cs" Link="TableListBuilder.cs" />
+    <Compile Include="..\RevEng.Core\TableListBuilder.cs" Link="TableListBuilder.cs" />
+    <Compile Include="..\RevEng.Core\Tables\CSharpModelGeneratorExtend.cs" Link="Tables\CSharpModelGeneratorExtend.cs" />
 	</ItemGroup>
 
   <ItemGroup>
--- a/src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
+++ b/src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
@@ -1,52 +1,28 @@
 ﻿using Microsoft.EntityFrameworkCore.Diagnostics;
 using Microsoft.EntityFrameworkCore.Metadata;
-using Microsoft.EntityFrameworkCore.Scaffolding;
-using Microsoft.EntityFrameworkCore.Scaffolding.Internal;
 using RevEng.Shared;
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 
-namespace RevEng.Core.Tables
+namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
 {
-    /// <summary>
-    ///     This is an extension of internal API that supports the Entity Framework Core infrastructure and not subject to
-    ///     the same compatibility standards as public APIs. The base CSharpModelGenerator may be changed or removed without 
-    ///     notice in any release. Updating Entity Framework Core version in project may result application failures.
-    ///     
-    ///     Base class implementation: https://github.com/dotnet/efcore/blob/main/src/EFCore.Design/Scaffolding/Internal/CSharpModelGenerator.cs
-    /// </summary>
     public class CSharpModelGeneratorExtend : CSharpModelGenerator
     {
-        /// <summary>
-        ///     This is an extension of internal API that supports the Entity Framework Core infrastructure and not subject to
-        ///     the same compatibility standards as public APIs. The base CSharpModelGenerator may be changed or removed without 
-        ///     notice in any release. Updating Entity Framework Core version in project may result application failures.
-        /// </summary>
-        private ReverseEngineerCommandOptions ReverseEngineerCommandOptions { get; set; }
+        private bool UseSchemaFolders { get; set; }
 
-        /// <summary>
-        ///     This is an extension of internal API that supports the Entity Framework Core infrastructure and not subject to
-        ///     the same compatibility standards as public APIs. The base CSharpModelGenerator may be changed or removed without 
-        ///     notice in any release. Updating Entity Framework Core version in project may result application failures.
-        /// </summary>
         public CSharpModelGeneratorExtend(
             ModelCodeGeneratorDependencies dependencies,
             ICSharpDbContextGenerator cSharpDbContextGenerator,
             ICSharpEntityTypeGenerator cSharpEntityTypeGenerator,
-            ReverseEngineerCommandOptions reverseEngineerCommandOptions)
+            bool useSchemaFolders)
             : base(dependencies, cSharpDbContextGenerator, cSharpEntityTypeGenerator)
         {
-            ReverseEngineerCommandOptions = reverseEngineerCommandOptions;
+            UseSchemaFolders = useSchemaFolders;
         }
 
         private const string FileExtension = ".cs";
 
-        /// <summary>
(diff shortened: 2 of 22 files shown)


ErikEJ:
Fantastic! I will have a closer look at the weekend.

src/GUI/RevEng.Core.50/RevEng.Core.50.csproj
```diff
@@ -33,7 +33,8 @@
 		<Compile Include="..\RevEng.Core\ServiceProviderBuilder.cs" Link="ServiceProviderBuilder.cs" />
 		<Compile Include="..\RevEng.Core\SharedTypeExtensions.cs" Link="SharedTypeExtensions.cs" />
 		<Compile Include="..\RevEng.Core\SqlServerHelper.cs" Link="SqlServerHelper.cs" />
-		<Compile Include="..\RevEng.Core\TableListBuilder.cs" Link="TableListBuilder.cs" />
+    <Compile Include="..\RevEng.Core\TableListBuilder.cs" Link="TableListBuilder.cs" />
```
ErikEJ, line 36: nit: Indent

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
```diff
@@ -0,0 +1,115 @@
...
+
+                // output EntityType poco .cs file
+                var entityTypeFileName = entityType.Name + FileExtension;
+                resultingFiles.AdditionalFiles.Add(
+                    new ScaffoldedFile 
+                    {
+                        Path = ReverseEngineerCommandOptions.UseSchemaFolders
+                            ? Path.Combine(entityType.FindAnnotation("Relational:Schema")?.Value.ToString() ??
+                                           entityType.FindAnnotation("Relational:ViewSchema")?.Value.ToString() ??
+                                           "dbo", entityTypeFileName)
```
ErikEJ, line 106: Why this special casing for SQL Server? Seems not needed.

ErikEJ (changes requested):

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
eminmesic: This is the only way where you can get information about table/view from annotations. If you remove this part you will not have schema name for tables and views and it will work only for Stored Procedures.

ErikEJ: There is this (see CommentCSharpEntityTypeGenerator):
```
            var schema = entityType.GetSchema();
            var defaultSchema = entityType.Model.GetDefaultSchema();
```
But what I meant that there is SQL Server specific schema name ("dbo"), which should not be there. There are some database systems that do not even have schemas, and that should just fall back to entitypeFileName.

I think this could be simplified to:
```
var schema = entityType.GetSchema();
Path = ReverseEngineerCommandOptions.UseSchemaFolders && !string.IsNullOrEmpty(schema) 
? Path.Combine(schema, entityTypeFileName)
: entitypeFileName

```

src/GUI/RevEng.Core.50/RevEng.Core.50.csproj
eminmesic: It is very strange, I don't see this indenting issue locally.

ErikEJ: Never mind..

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// <summary>
+    ///     This is an extension of internal API that supports the Entity Framework Core infrastructure and not subject to
+    ///     the same compatibility standards as public APIs. The base CSharpModelGenerator may be changed or removed without 
+    ///     notice in any release. Updating Entity Framework Core version in project may result application failures.
+    ///     
+    ///     Base class implementation: https://github.com/dotnet/efcore/blob/main/src/EFCore.Design/Scaffolding/Internal/CSharpModelGenerator.cs
+    /// </summary>
+    public class CSharpModelGeneratorExtend : CSharpModelGenerator
+    {
+        /// <summary>
```
ErikEJ, line 21: XML Summary not needed

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
```diff
@@ -0,0 +1,115 @@
...
+        ///     This is an extension of internal API that supports the Entity Framework Core infrastructure and not subject to
+        ///     the same compatibility standards as public APIs. The base CSharpModelGenerator may be changed or removed without 
+        ///     notice in any release. Updating Entity Framework Core version in project may result application failures.
+        /// </summary>
+        private ReverseEngineerCommandOptions ReverseEngineerCommandOptions { get; set; }
```
ErikEJ, line 26: Just " bool UseSchemaFolders´"

ErikEJ (changes requested):

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
ErikEJ: I wonder if this could simply be done here: https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/RevEng.Core/ReverseEngineerScaffolder.cs#L220 (so no need to override EF Core class)

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
ErikEJ: I wonder if this could simply be done here: https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/RevEng.Core/ReverseEngineerScaffolder.cs#L220 (so no need to override EF Core class)

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
eminmesic: Removed in the latest commit.

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
eminmesic: Replaced ReverseEngineerCommandOptions object with UseSchemaFolders property in dependency injection in the latest commit.

src/GUI/RevEng.Core/Tables/CSharpModelGeneratorExtend.cs
eminmesic: Improved this part in the latest commit `entityType.GetSchema()` returns dbo value for the views after EF Core 5.0 even you are using some specific schema. For that purpose, you have to call `entityType.GetViewSchema()` in the versions EF Core 5.0 and 6.0. You can see changes in the latest commit.

eminmesic: Regarding your message to try to not override the EF Core class, I would be happy if we can do that. However, in the line [https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/RevEng.Core/ReverseEngineerScaffolder.cs#L220](https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/RevEng.Core/ReverseEngineerScaffolder.cs#L220) we should iterate again through the model and split the directory from the file in the path to put schema information, which can be a little bit insecure.

As I said it would be better to not override the EF Core class because as you can see they still have some TO DO comments in the core one.

ErikEJ: Yes, but stored procedure results are not part of this use case.

ErikEJ: I will try to remove the override when this has been merged (or you could do it in a follow up PR?)

ErikEJ (approved):
Amazing - oldest issue finally fixed, more than 2 years old!

PR #1011, opened 2021-06-04
Model table-valued functions

### LINKED ISSUE #1006: List table-valued UDFs
Currently SPROCs and scalar functions are listed and can be reverse engineered. It would be nice to include table-valued functions into the mix.

### Steps to reproduce
At screen to select objects, table-valued functions are missing.

### Further technical details
EF Core Power Tools version: 2.5.665

Database engine: SQL Server

Visual Studio version: Visual Studio 2019 16.10

MisinformedDNA:
Fixes #1006


MisinformedDNA:
I could probably test it some more, but I'm out of time for the week. I'll pick back up on Monday if there are any issues, suggestions, etc.

ErikEJ:
Great, I will have a look start of next week, no time this weekend.

ErikEJ:
Would be great if you could add a TVF to the test script, and run code generation on the ScaffoldingTester project, so we can see the impact.

MisinformedDNA:
@ErikEJ I add test functions to the test script, `nwscript.sql`, but I did not commit the generated code since `CodeGenerationModel` is set to 1 (i.e.. EF 3.1) in the config file. How would you like me to proceed?

Also, there are still at least two problems with the current state:

1. All function result models are not set up in the `OnModelCreating`, meaning the user needs to add a line for each result class:

    ```csharp
    modelBuilder.Entity<SomeFunctionResult>().HasNoKey();
    ````

2. Each `decimal` type of the function result models issue a of warnings at runtime:

    > warn: Microsoft.EntityFrameworkCore.Model.Validation[30000]
      No type was specified for the decimal property 'QuantityPicked' on entity type 'SomeFunctionResult'. This will cause values to be silently truncated if they do not fit in the default precision and scale. Explicitly specify the SQL server column type that can accommodate all the values in 'OnModelCreating' using 'HasColumnType()', specify precision and scale using 'HasPrecision()' or configure a value converter using 'HasConversion()'.

I can work on these, but it'll probably take me a few more days.

ErikEJ:
I will check out the PR, and run through it.

I just have a few comments, but have not had time to go through it yet.

Maybe it is time to update to 5.0 for the test project.

Re 1: you should be able to add that to the .Functions partial DbContext.

Re 2: You should get the scale and precision from metadata, and maybe set in code. We can postpone that for second round.

MisinformedDNA:
> Re 1: you should be able to add that to the .Functions partial DbContext

But that would require implementing `OnModelCreatingPartial`, right? And that would break anyone who already has a custom implementation. Or am I missing something?

ErikEJ:
Hmmm... You are right. 

Let me have a closer look, and give it some thought.

ErikEJ:
We could inject a OnModelCreatingFunctionMappings method into the generated main DbContext, and then implement this in the partial Functions DbContrxt to add the entities?

MisinformedDNA:
And how do we inject into the main DbContext?

I see the context file is generated here

```csharp
var codeModel = codeGenerator.GenerateModel(model, codeOptions);
```

It looks like most of the generation is internal. Is there somewhere I can alter OnModelCreating before or during `GenerateModel` or should I just alter `codeModel.ContextFile` after `GenerateModel` completes, presumably searching for the right line and inserting a new line?

ErikEJ:
Add a new method to be called if Table Valued Functions are selected around here: https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/RevEng.Core/ReverseEngineerRunner.cs#L93 - and yes, simple string manipulation, sadly.

ErikEJ:
Getting closer to being ready?

MisinformedDNA:
I believe it is ready!

ErikEJ (approved):

ErikEJ:
Thanks, great contribution!

PR #1107, opened 2021-09-19
Bugfix for stored procedure return values

TonyMajorDev:
Bugfix to address stored procedure return values.  Previous version would return number of affected rows instead of the return value.  The return value will now be returned by a call from the generated Stored Procedure wrapper.


ErikEJ:
Please create a related issue, that demonstrates the bug. I am not aware of any bugs in this area.

ErikEJ:
Ping?

ErikEJ:
Returning affected rows is by design, to get the return value, pass in a OutputParameter to the method call.

PR #1139, opened 2021-10-14
Add ColumnTypeMappings option for custom column type mapping

jinweijie:
Added the ColumnTypeMappings option to allow custom CLR mapping to database column type.

### Usage Example

Example scenario: Map the database column type `timestamp with time zone` to DateTimeOffset for Postgres in reverse engineer.

In `efpt.config.json` file, add the following configuration:

```
  "ColumnTypeMappings": [
    {
      "StoreType": "timestamp with time zone",
      "ClrType": "System.DateTimeOffset"
    }
  ]
```
It will generate the code like:

`public DateTimeOffset SomeField { get; set; }` instead of `public DateTime SomeField { get; set; }` (the default one)

If the ClrType is not found, an exception will be thrown to VS Output window to let developer know which type was not found.


ErikEJ:
Given that this can be achieved in other ways and the very low number of requests for this, using for example a partial class for the entity, this is not something that I plan to add and support.

I will revisit if demand arises.

PR #1249, opened 2022-01-18
Support geo types as sproc parameters

### LINKED ISSUE #1247: Stored procedure with Geography type causes exceptions when executed
I am getting type mapping exceptions when running a stored procedure with a geography data type defined as a parameter.

The Power Tool generates the stored procedure method with the following parameter defined.
```
 new SqlParameter
  {
      ParameterName = "Location",
      Value = Location ?? Convert.DBNull,
      SqlDbType = System.Data.SqlDbType.Variant,
  },
```

When I call my function (`context.Procedures.stpInsertLocationAsync(...)`) I get the following exception if I have data for the point / Location parameter.

System.AggregateException: 'One or more errors occurred. (No mapping exists from object type NetTopologySuite.Geometries.Point to a known managed provider native type.)'
ArgumentException: No mapping exists from object type NetTopologySuite.Geometries.Point to a known managed provider native type.

And I get the following exception if the point / Location parameter is null.

System.AggregateException: 'One or more errors occurred. (Operand type clash: sql_variant is incompatible with geography)'
SqlException: Operand type clash: sql_variant is incompatible with geography

ErikEJ:
fixes #1247

### DIFF
--- a/src/GUI/RevEng.Core/Procedures/SqlServerSqlTypeExtensions.cs
+++ b/src/GUI/RevEng.Core/Procedures/SqlServerSqlTypeExtensions.cs
@@ -41,8 +41,8 @@ namespace RevEng.Core
                 { "rowversion", "timestamp" },
                 { "table type", "structured" },
                 { "sql_variant", "variant" },
-                { "geography", "variant" },
-                { "geometry", "variant" },
+                { "geography", "udt" },
+                { "geometry", "udt" },
                 { "hierarchyid", "variant" },
                 { "sysname", "nvarchar" },
             });
@@ -157,8 +157,10 @@ namespace RevEng.Core
                     return isNullable ? typeof(byte?) : typeof(byte);
 
                 case SqlDbType.Variant:
-                case SqlDbType.Udt:
                     return typeof(object);
+
+                case SqlDbType.Udt:
+                    return typeof(byte[]);
 
                 case SqlDbType.Structured:
                     return typeof(DataTable);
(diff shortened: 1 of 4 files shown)


nathanh2011:
Hi Erik,

I just checked your changes and there is a slight error. When creating the parameters and setting the UDT type, you have used the property 'TypeName' instead of 'UdtTypeName'.

So this:

```
new SqlParameter
  {
      ParameterName = "Geo1",
      Value = Geo1 ?? Convert.DBNull,
      SqlDbType = System.Data.SqlDbType.Udt,
      TypeName = "geography",
  }
```

should be this:

```
new SqlParameter
  {
      ParameterName = "Geo1",
      Value = Geo1 ?? Convert.DBNull,
      SqlDbType = System.Data.SqlDbType.Udt,
      UdtTypeName = "geography",
  }
```

Also, is there anyway to have the function accept a geography type for the parameter and convert it to bytes inside the function, so the caller knows it is meant to be a geography and no accidently pass a random byte array? I am not sure if that is possible in a clean way but just thought I would ask.

But thanks for sorting this out so quickly.

ErikEJ:
Thanks, will fix asap.

What  "geography type" are you thinking of - as far as I know there is no such type in .NET Core, only in a 3rd party package.

ErikEJ:
@nathanh2011  Fixed in latest build

nathanh2011:
Sorry, I am getting my geometry and geographies confused.

The function could use the Geometry from the NetTopologySuite package to represent a SQL Geography type, giiven that is what Microsoft seem to be suggesting to use (https://docs.microsoft.com/en-us/ef/core/modeling/spatial). 

It would be interesting to see how EF Core actually saves/loads and converts geography data into NTS geometry through a normal class and Context.Save() and maybe use a similar mechanism. I tried to look but have no idea where to find that functionality.

ErikEJ:
They do more or less the same:
https://github.com/dotnet/efcore/blob/main/src/EFCore.SqlServer.NTS/Storage/Internal/SqlServerGeometryTypeMapping.cs#L146

https://github.com/dotnet/efcore/blob/main/src/EFCore.SqlServer.NTS/Storage/ValueConversion/Internal/GeometryValueConverter.cs#L25

Not sure how that could be added to the generated code, but if you figure something general out, feel free to share

Issue #1367, opened 2022-05-02
Reverse Engineer - Suggest some Default Values


doggy8088:
I have some suggestions to discuss.

1. EF Core 5 is the default now. I'd like to suggest change it to EF Core 6 by default.

    ![image](https://user-images.githubusercontent.com/88981/166241205-28cca27d-3ee1-48e1-90d9-883241954614.png)

2. The `EntityTypes path` and `DbContext path` can fill default value directly and remove `(f.ex. xxxx)` part.

    ![image](https://user-images.githubusercontent.com/88981/166241684-b47eb23e-bccc-478e-b997-a8ca14f4f21a.png)

3. See above screenshot. The `Context name` shown `ContosouniversityContext` that extract from my connection string. My database name is actually called `ContosoUniversity`. Is it possible leave database name as is just like `ContosoUniversityContext`?

ErikEJ:
Fine to default to EF6 - PR?

Not so keen on item 2 - too breaking and opinionated.

On 3: The DbContext names is C#-ified, so maybe it loses some casing details. Have a look to see if it can be improved.

doggy8088:
1. Ok. I'll do it in the coming days.
2. That's just a "default value". I think "opinionated" is fine. Better than empty.
3. If you use `dotnet ef dbcontext scaffold`, the generated dbcontext name will be `ContosoUniversityContext`.

ErikEJ:
I suggest creating individual issues instead of piling them up.

I have created a new issue for item 3.

doggy8088:
OK. It's good.

PR #1371, opened 2022-05-02
Move EF Core 6 to the first item #1367

doggy8088:


ErikEJ (changes requested):
That wont work. It is serialized as int.

ErikEJ:
I think this line should set a new default: https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/Shared/Handlers/ReverseEngineer/ReverseEngineerOptions.cs#L50

doggy8088:
done.

ErikEJ:
Thanks a lot!

PR #1563, opened 2022-11-01
first check in of nav renaming code. Compiles but not tested yet

R4ND3LL:
Here is the nav renaming code.    Rider auto-formatted a couple files here creating some formatting changes -- sorry bout that.

Aside from the obvious new code, note the important addition of ColumnRenamer.AlternateName which i use to suggest the other possible nav name based on the 2 styles generated.


ErikEJ:
I cannot take this in, breaks too many existing code analyzer rules.

I have pulled out the main parts, and will do a PR on those tomorrow, that you can review (and then test 😄 )

PR #1565, opened 2022-11-01
Enable renaming of navigation and reference properties

### LINKED ISSUE #1562: Renaming Navigations Design Proposal
@R4ND3LL please submit a draft PR!

_Originally posted by @ErikEJ in https://github.com/ErikEJ/EFCorePowerTools/issues/211#issuecomment-1297642675_

@ErikEJ, I have the fork/branch created.  I don't have a whole lot of time to devote to this; Can you give me a couple of suggestions as to where this code would be best added into your solution?      

This feature requires a couple of nuget packages (like Microsoft.CodeAnalysis.CSharp.Workspaces).  It will look for the new renaming-nav json file, load it, and process the renames on the already-generated code.

There are a couple of options for WHEN this can take place:
- At end of "Reverse Engineer" process (my preference)
- Manually invoked by user via a new menu option.

Let me know what you think.

ErikEJ:
related to #211

fixes #1562

### DIFF
--- a/src/GUI/Shared/Extensions/ProjectExtensions.cs
+++ b/src/GUI/Shared/Extensions/ProjectExtensions.cs
@@ -96,7 +96,7 @@ namespace EFCorePowerTools.Extensions
             return result.OrderBy(s => s).ToList();
         }
 
-        public static string GetRenamingPath(this Project project, string optionsPath)
+        public static string GetRenamingPath(this Project project, string optionsPath, bool navigationsFile = false)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
@@ -116,7 +116,9 @@ namespace EFCorePowerTools.Extensions
                 renamingPath = Path.GetDirectoryName(optionsPath);
             }
 
-            return Path.Combine(renamingPath, "efpt.renaming.json");
+            const string efptRenamingJson = "efpt.renaming.json";
+            const string efptRenamingNavJson = "efpt.refrenaming.json";
+            return Path.Combine(renamingPath, navigationsFile ? efptRenamingNavJson : efptRenamingJson);
         }
 
         public static async Task<Tuple<bool, string>> ContainsEfCoreReferenceAsync(this Project project, DatabaseType dbType)
--- a/src/GUI/Shared/Helpers/RoslynEntityPropertyRenamer.cs
+++ b/src/GUI/Shared/Helpers/RoslynEntityPropertyRenamer.cs
@@ -1,0 +1,146 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+
+namespace EFCorePowerTools.Helpers
+{
+    internal static class RoslynEntityPropertyRenamer
+    {
+        /// <summary> apply renaming using Roslyn, return rename count.</summary>
+        /// <param name="path">path to renaming rules file. can process primitive or navigation rules.</param>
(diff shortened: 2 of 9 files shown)


src/GUI/Shared/Extensions/ProjectExtensions.cs
```diff
@@ -116,7 +116,9 @@ public static string GetRenamingPath(this Project project, string optionsPath)
                 renamingPath = Path.GetDirectoryName(optionsPath);
             }
 
-            return Path.Combine(renamingPath, "efpt.renaming.json");
+            const string efptRenamingJson = "efpt.renaming.json";
+            const string efptRenamingNavJson = "efpt.refrenaming.json";
```
ErikEJ, line 120: @R4ND3LL Notice new file name

src/GUI/Shared/Helpers/RoslynEntityPropertyRenamer.cs
```diff
@@ -0,0 +1,146 @@
...
+                }
+                catch
+                {
+                    // TODO respond to this from caller
+                    return -1; // log this?
+                }
+            }
+
+            var rulesText = File.ReadAllText(path);
+            var rulesList = RenamingRulesSerializer.TryRead(rulesText);
```
R4ND3LL, line 91: TryRead is expecting a path here.  Pass "path" instead of "rulesText" and remove previous line

R4ND3LL (changes requested):
I made one comment about the usage of TryRead in RoslynEntityPropertyRenamer.   this should be updated

ErikEJ:
I think it is a bit odd to use names like "Table" and "Column" in the json file here, but I will try to merge this now, after some smoke testing, and then consider once it has beeen tested more.

ErikEJ:
Instead of this:

```json
[
  {
    "SchemaName": "dbo",
    "Tables": [
      {
        "Columns": [
          {
            "Name": "Picture",
            "NewName": "Image"
          }
        ],
        "Name": "Categories",
        "NewName": "Categories"
      }
    ],
    "UseSchemaName": false
  }
]
```

something like this?

```json
[
  {
    "Class": [
      {
        "Properties": [
          {
            "Name": "Picture",
            "NewName": "Image",
            "Alternatename": "abc"
          }
        ],
        "Name": "Categories"
      }
    ]
  }
]

```

R4ND3LL:
Yes, that makes sense (using class/properties terminology).   Would just need a separate model.

To be a little pedantic, the file name "efpt.refrenaming.json" kind of suggests that it will only rename "references" as opposed to using the term nav or navigations, which includes references & collections.   

To take that a step further, technically this file can accomplish ANY property renaming so if someone wanted to use this file alone they could include all properties in it.  From that perspective, the first rule file is more of an "in-generation level renaming" whereas the new file is "post-generation renaming".   You could use that concept in the name of the file instead of "ref" (ex. post-gen-renaming).

Do you intend to move this to a command line-level library?  Hit a roadblock with that?

ErikEJ:
Please consider all this experimental at the moment...

>> the file name "efpt.refrenaming.json" kind of suggests that it will only rename "references"

Yes, a new model and a better name - maybe "efpt.propertyrenaming.json"?

>> Do you intend to move this to a command line-level library? Hit a roadblock with that?

Ideally, yes, but this will require the package install to also be executed by the command line tool. 
Which could cause headaches (it is not aware of the VS package sources)

I have created https://github.com/ErikEJ/EFCorePowerTools/issues/1567

Looking forward to hearing more testing feedback from you!

R4ND3LL:
Test passed.  Renamed nav properties in target classes as well as Configuration references.
Project builds.. it all looks good

R4ND3LL:
"efpt.property-renaming.json" is good (with hyphenation to make it more readable like "dotnet-tools.json")

PR #1568, opened 2022-11-03
Get status from renamer

ErikEJ:
Always use UTF8
Update test project


ErikEJ:
@R4ND3LL FYI (notice use of UTF8 encoding always)

PR #1571, opened 2022-11-04
EdmxRuleGenerator done

R4ND3LL:
EdmxRuleGenerator done.  Generates 3 rule json files including column renames, nav property renames, and enum mapping json.
changed both generator and renamer projects to exe, with smart arg inspection for easy debugging.
Minor refactor of rule cs file locations.  

Note, once the code is accepted on your end, I'll update the renamer project to process the enum rules using Roslyn.  
Then we can look at doing a UI for the generator in the VSIX.

Also I can try cleaning up the code style afterwards too, it's just time-consuming for now...


ErikEJ:
I am reluctant to accept this huge amount of cider that targets a very narrow use case.

Maybe ir would be better if you simple package abd publish this as a dotnet tool. Happy to help with that.

R4ND3LL:
It’s close to providing complete edmx support for upgraders. That’s not such a narrow use case IMO. 
The EDMX model is what it is. It might be large but at least it’s final and needs no maintenance. 
Give the changes a read, I think you’ll find it pretty straight forward.
In any case, let me know what you want to do.
I could potentially deliver it another way but I didn’t want to invest a ton of time on it; your VSIX is a smart location for it.

ErikEJ:
I have only had 2 users ask for soemthing like this in the lifetime of this extension. (and you are one of them)

I suggest a dotnet tool, and I can provide a menu item with a link to a wiki page explaing the installation and usage.

ErikEJ:
Did you actually get the RoslynRenamer to work?

R4ND3LL:
> Did you actually get the RoslynRenamer to work?

Yes.  Turned the renamer project into an exe to test invoking directly; project loading works with no problem when targeting 6.0.  As you said, since the CLI is not supported anyway, and always works consistently via VSIX, it's kind of a non-issue.  There may be some workarounds to explore though.

This code is tested on 11 EDMX files so far and correctly generates rule files, including your original renaming.json file, and applies the name changes with no problem.  Basically, the user has to edit nothing by hand in order to migrate from EF6.  

When researching your repo before I started, I got the impression that there was interest in EDMX support.  Found over 30 issues referencing EDMX and over 100 referencing EF6 -- so I know people would use this if they knew it was available.  Current literature tells everyone that EDMX is just not possible in Core. 

What are your reservations about using the code?  If it's just from a maintenance perspective; if these libraries were NuGet packages managed by me, would you consider referencing in the VSIX and providing the same functionality in-line?

I admit, I selfishly want this built into the VSIX because I like the UI-driven easy approach to scaffolding for my team.  Simplifies and saves time.  As opposed to breaking scaffolding up into multiple steps.

ErikEJ:
Ah! Maybe targrting 3.1 would be my issue.

Yes, I have other "fringe" features and they are all based on Nuget packages maintained elsewhere so that would be ideal. Just need to decide on the API surface.

R4ND3LL:
I can combine the EdmxRuleGen and RoslynRenamer projects into one (nuget package).   

In order to generate the rule files, I would take ownership of the json models, including your original renaming.json model.  Otherwise all the primitive property rules still need to be defined by hand.

We still need to tackle the simple form in the VSIX that is used to select an EDMX and call the API to generate the rules.  This form would likely need to live in your project.    

If you're ok with both of the above, we can pivot to that plan.

ErikEJ:
>> Basically, the user has to edit nothing by hand in order to migrate from EF6.

That sounds very compelling!

ErikEJ:
>> In order to generate the rule files, I would take ownership of the json models, including your original renaming.json model. Otherwise all the primitive property rules still need to be defined by hand.

Fine by me, if they diverge I can always convert. 

So your nuget Lib would take the path to an EDMX and return two strings?

R4ND3LL:
> > > Basically, the user has to edit nothing by hand in order to migrate from EF6.
> 
> That sounds very compelling!

Honestly, I think the community was shocked by the lack of an upgrade path for EF6 EDMX projects.  This will be the first and only end-to-end automated solution for upgrading.   I expect it to gather a lot of praise once fully delivered.

I don't think anyone has approached the problem like this before.  Generate rules from the EDMX, then discard the EDMX.  Apply rules on scaffolding.   I'm highly confident it will work well.

ErikEJ:
Yeah, sounds quite promising/ awesome. As you can see there was a failed attempt earlier!

R4ND3LL:
> So your nuget Lib would take the path to an EDMX and return two strings?

I'm thinking the lib would take the path to an EDMX and return 3 object models:
1. primitive table/column naming rules
2. navigation naming rules
3. enum type - property mapping (this is currently working in the PR too)

the caller can then decide what to do with the models. in your case, serialize to json files.

ErikEJ:
Sounds great! I am going to close this - I will of course take care of all the wiring up in a good location - maybe just as an advanced option during reverse engineering (an optional path to the .edmx file,)

PR #1664, opened 2023-01-19
Add Arm64 support for Reverse Engineering

DavidWhelan1890:
I modified the build files to also generate the reverse engineering console apps for arm64. When the extension loads on of them it determines what the cpu architecture it is on and extracts the correct one.

This is not the most elegant solution by any means but it leaves all of the original x64 build instructions and steps as untouched as possible

### DIFF
--- a/src/GUI/EFCorePowerTools/EFCorePowerTools.csproj
+++ b/src/GUI/EFCorePowerTools/EFCorePowerTools.csproj
@@ -174,12 +174,20 @@
       <Link>efreveng60.exe.zip</Link>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
+    <Content Include="..\lib\efreveng60arm.exe.zip">
+      <Link>efreveng60arm.exe.zip</Link>
+      <IncludeInVSIX>true</IncludeInVSIX>
+    </Content>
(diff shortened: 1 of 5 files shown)


DavidWhelan1890:
Forgot to add. It has been tested as well as it could. It loaded, connected to a db, and generated the models and context. Unfortunately sql server express does not run on arm so it has not been tested exhaustively

ErikEJ:
Cool, I will have a look, but seems a visble approach.

ErikEJ:
Maybe SQL Edge Container?

DavidWhelan1890:
That was some of my concern as well. I have been thinking on some other ways to solve the problem but haven't come up with anything as of yet

Just looked at sql edge containers and it looks like that might be a workable solution. When I get a change I will load one up and report the results

ErikEJ:
One approach to reducing foot print could be to download from somewhere, for example publish the command line tool as an (unsupported) global .net tool on NuGet. But then you will run into many other issues.

DavidWhelan1890:
I installed it fresh and it looks to be working exactly as it should! I will definitely leave a rating/review on the marketplace

ErikEJ:
Excellent - let's live with the slight increase in payload then

PR #1691, opened 2023-02-14
This might be bug where space was missing but this time before -v

meetkerasiya:


ErikEJ:
No a bug, there is a space after {packagId}

PR #1780, opened 2023-04-27
fix(cli): get the full path of efpt.renaming.json which is at the roo…

ng-julien:
efpt.renaming.json is at the root of the project, but this one is not taken into account

### DIFF
--- a/src/GUI/RevEng.Shared/Efcpt/EfcptConfigMapper.cs
+++ b/src/GUI/RevEng.Shared/Efcpt/EfcptConfigMapper.cs
@@ -67,7 +67,7 @@ namespace RevEng.Common.Efcpt
                 IncludeConnectionString = config.codegeneration.enableonconfiguring,
                 SelectedToBeGenerated = selectedToBegenerated,
                 Dacpac = isDacpac ? connectionString : null,
-                CustomReplacers = GetNamingOptions(projectPath),
+                CustomReplacers = GetNamingOptions(),
                 UseLegacyPluralizer = config.codegeneration.uselegacyinflector,
                 UncountableWords = config.replacements?.uncountablewords.ToList(),
                 UseSpatial = config.typemappings?.usespatial ?? false,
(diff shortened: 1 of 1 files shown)


ErikEJ:
Hmm.. The assumption is that the efpt.renaming.json file lives in the same folder a efcpt-config.json  - not sure that this PR cvers that (agree that current implementation is afulty)

ErikEJ:
Maybe I also need an option to override the path assumption for the renaming file?

ng-julien:
The problem occurs if optionsPath = "c:\myFolder\mySolution\src\myProject";
var optionsCustomNamePath = Path.Combine(Path.GetDirectoryName(optionsPath), "efpt.renaming.json"); 
The value of optionsCustomNamePath is c:\myfolder\mysolution\src\efpt.renaming.json
but I think the value should be  c:\myfolder\mysolution\src\myProject\efpt.renaming.json

maybe I'm wrong?

ErikEJ:
@ng-julien You are correct

ng-julien:
the path of efcpt-config.json is c:\myFolder\mySolution\src\myProject

ng-julien:
in my terminal:
i'm at  c:\myFolder\mySolution\src\myProject
the command efcpt "Data Source=localhost;Initial Catalog=MyCatalog;User Id=MyUserId;Password=MyPassword;"  Microsoft.EntityFrameworkCore.SqlServer

config file and renamling file are in same folder "myProject"

ErikEJ:
@ng-julien I have updated you fix to also respect a custom path to the config file

ErikEJ:
@ng-julien  Could you get the latest master and verify my update, then I will publish preview 3

ErikEJ:
@ng-julien Fix is here: https://github.com/ErikEJ/EFCorePowerTools/commit/758d5aabdfe6affc49d141c14e868a8ef8f11c7c

PR #1811, opened 2023-05-21
Use IAsyncEnumerable in DbContextExtensions

superrnovae:
This PR changes the return type of `SqlQuery` methods from `Task<List<T>>` to `IAsyncEnumerable<T>` so that only one object at a time is loaded into memory.

Note: I haven't tested it, as I'm having issues building the project on VS2022 and VS2019.
I also haven't touched DbContextExtensions text files in `RevEng.Core60` and `RevEng.Core70` projects.

Once above is done, I believe `SqlServerStoredProcedureScaffolder.cs` would also need to be fixed at the line 527.

### DIFF
--- a/src/GUI/ErikEJ.EntityFrameworkCore.6.SqlServer.SqlQuery/DbContextExtensions.cs
+++ b/src/GUI/ErikEJ.EntityFrameworkCore.6.SqlServer.SqlQuery/DbContextExtensions.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Data.Common;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


ErikEJ:
Can you build the NuGet project? What build errors are you getting? The solution requires VS 2022 and the Extensibility workload

superrnovae:
> Can you build the NuGet project? What build errors are you getting? The solution requires VS 2022 and the Extensibility workload

I'm getting this error: `Extension 'f4c4712c-ceae-4803-8e52-0e2049d5de9f' could not be found. Please make sure the extension has been installed` and a warning `The extension is incompatible with the targeted version of Visual Studio. More info at https://aka.ms/ExtensionSdkErrors. Following references are incompatible. Microsoft.VisualStudio.Web.BrowserLink.12.0.dll -> envdte`.

I'm on a VS2022 Professional LTSC, with Extensibility workload installed.

ErikEJ:
The warning you can ignore. In which context do you get the error?

superrnovae:
> In which context do you get the error?

When I'm trying to build the EFCorePowerTools project.

ErikEJ:
This PR will not have any effect on how code is generated in Power Tools. 

I am also concerned about the breaking change effect of this.

You are also the first user to see this (ToList) as a concern.

ErikEJ:
I think this should be factored as something you can opt-in to via an global options, similar to opt-in for non-async usage.

PR #2038, opened 2023-12-10
Zh hant translation

ErikEJ:
Based on #2037  from @doggy8088 - with some .csproj updates


ErikEJ:
@doggy8088 I did some required  .csproj updates

PR #2214, opened 2024-03-08
Added MergeDacpac property from Config in initial CLI analyitics

marc577:
I had the Issue, that I don't want the DacPacs to be Merged at the end. So i declared the property in the JSON Config. But in the initial analytics, the dacpacs will be merged together, even if the property is set to false.

So i thought the way to go is to get this from the config file.

Please give it a review :)


ErikEJ:
This does not seem necessary...

ErikEJ:
I fixed this in the latest nightly - see https://github.com/ErikEJ/EFCorePowerTools/commit/e3a4738c83b48ee5e71c8758a8d2d35919a3c3bc

PR #2221, opened 2024-03-13
add a global.json

SimonCropp:
so we can ensure a reliable and consistent sdk version

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -14,6 +14,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - uses: actions/setup-dotnet@v4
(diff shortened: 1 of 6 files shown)


ErikEJ:
Looks like the pipeline must be updated?

ErikEJ:
How is this different from letting the build agent just use the latest stable?

SimonCropp:
* for me i want to know that local, on CI, and anyone contributing. all are using the latest sdk. so i can rule out any bugs in older sdk as a culprit in diagnosing issues.
 * sometimes perf (and other) bugs are fixed in the compiler (hence sdk). so i want all my projects to be ensured of having these fixes

ErikEJ:
Thanks!

PR #2262, opened 2024-04-04
fix: #2261 set nullable annotation for procedure files

### LINKED ISSUE #2261: #nullable -annotations not set for procedures
### Provide steps to reproduce a bug

Reverse engineering a SQL Server database with procedures generates the `#nullable` annotations to the file start as expected for the entities, but the `IContextProcedures` interface or the implementation do not have the annotations. This leads to type confusion when the application at large expects nullable references to work.

Is this something to be added to WriteDbContextInterface/WriteDbContext in SqlServerStoredProcedureScaffolder? 

### Provide technical details

- EF Core Power Tools version: 2.6.200

- Exact Visual Studio version: Visual Studio 2022 Version 17.8.6

- Database engine: SQL Server

- Is T4 templates used: no

spaasis:
Since the procedure parameters are always nullable, set the `#nullable disable` annotation to the relevant files to reduce conflicts with projects using nullable reference types

Tested with the CLI - generates the expected annotations to the top of the two files - the interface definition and the implementation:

```
// <auto-generated> This file has been auto generated by EF Core Power Tools. </auto-generated>
#nullable disable
using Microsoft.Data.SqlClient;
...
```

fixes #2261

### DIFF
--- a/src/Core/RevEng.Core.60/Procedures/SqlServerStoredProcedureScaffolder.cs
+++ b/src/Core/RevEng.Core.60/Procedures/SqlServerStoredProcedureScaffolder.cs
@@ -52,6 +52,7 @@ namespace RevEng.Core.Procedures
             Sb = new IndentedStringBuilder();
 
             Sb.AppendLine(PathHelper.Header);
+            Sb.AppendLine("#nullable disable"); // procedure parameters are always nullable
 
             var usings = CreateUsings(scaffolderOptions, model, schemas);
 
@@ -158,6 +159,7 @@ namespace RevEng.Core.Procedures
             Sb = new IndentedStringBuilder();
 
             Sb.AppendLine(PathHelper.Header);
+            Sb.AppendLine("#nullable disable"); // procedure parameters are always nullable
 
             var usings = CreateUsings(scaffolderOptions, model, schemas);
 
--- a/src/Core/RevEng.Core.60/Procedures/SqlServerStoredProcedureScaffolder.cs
+++ b/src/Core/RevEng.Core.60/Procedures/SqlServerStoredProcedureScaffolder.cs
@@ -52,11 +52,7 @@ namespace RevEng.Core.Procedures
             Sb = new IndentedStringBuilder();
 
             Sb.AppendLine(PathHelper.Header);
-
-            if (scaffolderOptions.NullableReferences)
-            {
-                Sb.AppendLine("#nullable enable");
-            }
+            Sb.AppendLine("#nullable disable"); // procedure parameters are always nullable
 
             var usings = CreateUsings(scaffolderOptions, model, schemas);
 
@@ -138,7 +134,7 @@ namespace RevEng.Core.Procedures
 
                 foreach (var procedure in model.Routines)
                 {
-                    GenerateProcedure(procedure, model, false, scaffolderOptions.UseAsyncCalls, scaffolderOptions.NullableReferences);
+                    GenerateProcedure(procedure, model, false, scaffolderOptions.UseAsyncCalls);
                 }
 
                 if (model.Routines.Exists(r => r.SupportsMultipleResultSet))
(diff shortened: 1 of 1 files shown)


ErikEJ:
Wonder if it would be better to actually support nullable reference types if opted in, this already happens in the result classes

spaasis:
@ErikEJ 

It surely would be more explicit. That would then require that all reference types are suffixed with `?`

So some kind of fiddling with this in GenerateProcedure should be enough?

```
 var paramStrings = procedure.Parameters.Where(p => !p.Output)
                .Select(p => $"{Code.Reference(p.ClrType(asMethodParameter: true))} {Code.Identifier(p.Name)}")
                .ToList();

```

ErikEJ:
I do soemthing similar for the result set here:

https://github.com/ErikEJ/EFCorePowerTools/blob/b94ef6b0f3250e894657d75af44242734157387b/src/Core/RevEng.Core.60/Routines/SqlServerRoutineScaffolder.cs#L362

(Also respecting options.UseNullableReferences)

ErikEJ:
and then add #nullable enable to the top instead

spaasis:
@ErikEJ could we add the handling as far down as SqlServerTypeExtensions? There's a method GetClrType which handles all types and already handles null for non-reference types. So would it be simpler to pass the config flag here and handle the parsing in one place?

ErikEJ:
> simpler to pass the config flag here and handle the parsing in one place?

Not sure I understand how that would append the "?" in code generation?

spaasis:
@ErikEJ I was thinking along these lines:

Currently that function does this:
```
  public static Type GetClrType(string storeType, bool isNullable, bool asParameter = false)
        {
            var sqlType = GetSqlDbType(storeType);

            var useDateOnlyTimeOnly = UseDateOnlyTimeOnly;

            switch (sqlType)
            {
                case SqlDbType.BigInt:
                    return isNullable ? typeof(long?) : typeof(long);
```

So could it work if we passed the config value as `nullableReferences` and did:
```
                case SqlDbType.Char:
                case SqlDbType.NChar:
                case SqlDbType.NText:
                case SqlDbType.NVarChar:
                case SqlDbType.Text:
                case SqlDbType.VarChar:
                case SqlDbType.Xml:
                    return nullableReferences && isNullable ? typeof(string?) : typeof(string);
```

And the same for the other reference types

spaasis:
And I just assume that is that function returns a `typeof(string?)` the resulting code generation is `string?` but maybe that's not the case..

ErikEJ:
You can try - there is also OutputParameter - using the two conditions 

options.UseNullableReference && !propertyType.IsValueType

seems like the better option to me.

spaasis:
Here's an implementation that modifies the `GenerateProcedure` and initial test seems okay, but I'll have to test more thoroughly tomorrow. If this approach suits you then I won't go down the rabbit hole of `GetClrType` which may have some serious implications elsewhere

ErikEJ:
Nice! (Sorry for the half baked implementation from my side!)

spaasis:
Ah, but when we're enabling nullable here there's some issues with the OutputParameters:
The OutputParameters itself is not implemented as nullable and it gives errors when the feature is enabled. If this is fixed:
```
public class OutputParameter<TValue> {
    private bool _valueSet = false;

    public TValue? _value;

    public TValue? Value {
        get {
            if (!_valueSet)
                throw new InvalidOperationException("Value not set.");

            return _value;
        }
    }

    internal void SetValue(object? value) {
        _valueSet = true;

        _value = null == value || Convert.IsDBNull(value) ? default(TValue) : (TValue)value;
    }
}
```


There are some issues with the current implementation:
First, for procs that return a single int return value the code is generated as `OutputParameter<int> returnValue = null` and of course the `= null` gives a warning here.

And second, the output parameters are handled in a way that causes type confusion:

1. Parameter definition `OutputParameter<int?> MyParam`
2. SqlParameter assignment has en excess `?` in the `Value` setter
```
    var parameterMyParam = new SqlParameter
            {
                ParameterName = "MyParam",
                Direction = System.Data.ParameterDirection.InputOutput,
                Value = MyParam?._value ?? Convert.DBNull,
                SqlDbType = System.Data.SqlDbType.Int,
            };
```
3. Return value assignment throws an error `MyParam.SetValue(parameterMyParam.Value);` since at this point the compiler believes the earlier `MyParam?._value´ and assumes that `MyParam` can be nullable here as well.


This is getting a bit hairy :p

ErikEJ:
OK - maybe I did not complete the implementation for a reason.

Let's test with just #disable then

PR #2478, opened 2024-08-27
fill DefaultValue for dacpac

m0a0k0s:
In ef8 there is DefaultValue property in addition to DefaultValueSql. EF scaffolder looks at this property and if it's not-empty for 
a bool type, the result field will be generated as non-nullable bool. Since it's not filling now, it is always generated as bool?, which is different from what is generated directly from the database.

I've taken TryParseClrDefault implementation from efcore repository
https://github.com/dotnet/efcore/blob/017fb22e4b63003dc650905ea066b992895dc54d/src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs#L843
https://github.com/dotnet/efcore/blob/017fb22e4b63003dc650905ea066b992895dc54d/src/Shared/SharedTypeExtensions.cs#L77

### DIFF
--- a/src/Core/NUnitTestCore/Dacpac/DacpacTest.cs
+++ b/src/Core/NUnitTestCore/Dacpac/DacpacTest.cs
@@ -60,7 +60,7 @@ namespace UnitTests
         public void CanEnumerateSelectedQuirkObjects()
         {
             // Arrange
-            var factory = new SqlServerDacpacDatabaseModelFactory(null, null);
+            var factory = new SqlServerDacpacDatabaseModelFactory(null);
             var tables = new List<string> { "[dbo].[FilteredIndexTable]", "[dbo].[DefaultComputedValues]" };
             var options = new DatabaseModelFactoryOptions(tables, new List<string>());
 
@@ -82,7 +82,7 @@ namespace UnitTests
         public void CanEnumerateSelectedComputed()
         {
             // Arrange
-            var factory = new SqlServerDacpacDatabaseModelFactory(null, null);
+            var factory = new SqlServerDacpacDatabaseModelFactory(null);
             var tables = new List<string> { "[dbo].[DefaultComputedValues]" };
             var options = new DatabaseModelFactoryOptions(tables, new List<string>());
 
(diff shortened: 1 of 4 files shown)


ErikEJ:
Looks good.

Could you address the build failures?

m0a0k0s:
updated, but there's an issue with sdk version in build pipeline

ErikEJ:
@m0a0k0s Would be grateful if you could update the tools for this, then you can use you changes in a daily build.

Run "src\Core\efreveng60\BuildCmdlineTool.cmd" 
Run "src\Core\efreveng70\BuildCmdlineTool.cmd" 
Run "src\Core\efreveng80\BuildCmdlineTool.cmd" 

and create a PR with the 3 updated .zip files.

ErikEJ:
There is also a build error, please build full solution locally for any future PRs:

https://github.com/ErikEJ/EFCorePowerTools/commit/00c818c95819d7ebaa20ab6eec93fb7f29940eba#annotation_25615351907

ErikEJ:
@m0a0k0s Did you manage to verify this fix?

m0a0k0s:
I use modified ErikEJ.EntityFrameworkCore.SqlServer.Dacpac nuget only with code from this PR, not the whole powertool. It works, but it not exactly the daily build.

ErikEJ:
@m0a0k0s I will get an updated nuget package published!

chassq:
Wanted to know if anyone has seen this issue. After updating to 2.6.577 our models changed where we had BIT fields in Azure SQL that are required and defult to 1 (true). The models now remove the nullable (?) on the property and the configuration went from 

```
 entity.Property(e => e.IsActive)
     .IsRequired()
     .HasDefaultValueSql("1");
```

to 

`entity.Property(e => e.IsActive).HasDefaultValue(true);`

We are testing this right now but is there an issue with this whereby the model would cause the value to default to false so we would push the value of false to the database as the default instead of what the database has specified that the default should be true? Does the configuration actually end up specifying the default value for a new model?

ErikEJ:
Sounds like you also moved to EF Core 8?? It is by design, the tool has a option to ignore your strange default.

chassq:
We are on EF Core 8. Could please point us to the option? We see the **Remove SQL default from bool columns**. Not sure if that is the one but if we select it or not we get the same result.

ErikEJ:
Have you tested if there is any negative effect?

ErikEJ:
Just noticed you highjack a closed issue. Please create a new issue with full reproduction details. Are you even using a .dacpac??

PR #2546, opened 2024-09-27
Resolves #2545 - Adds a warning to the output if both 'code-generation/use-database-names' is set to true and 'efpt.renaming.json' exists.

### LINKED ISSUE #2545: Renaming doesn't work if UseDatabaseNames is set to true, add warning if you try to use both
So it turns out that if you have `use-database-names` set to true, having a efpt.renaming.json file won't do you any good.  Since I had EFCPT building locally already, I wanted to figure out why.  

It successfully gets into https://github.com/ErikEJ/EFCorePowerTools/blob/2793e2ae457f8d765176479a51d1dcd6deccb468/src/Core/RevEng.Core.60/ServiceProviderBuilder.cs#L73 with the `options.CustomReplacers` passed in as expected.  But it would never hit breakpoints in  https://github.com/ErikEJ/EFCorePowerTools/blob/2793e2ae457f8d765176479a51d1dcd6deccb468/src/Core/RevEng.Core.60/ReplacingCandidateNamingService.cs#L37

and 
https://github.com/ErikEJ/EFCorePowerTools/blob/2793e2ae457f8d765176479a51d1dcd6deccb468/src/Core/RevEng.Core.60/ReplacingCandidateNamingService.cs#L92

It will hit for 

https://github.com/ErikEJ/EFCorePowerTools/blob/2793e2ae457f8d765176479a51d1dcd6deccb468/src/Core/RevEng.Core.60/ReplacingCandidateNamingService.cs#L147

so I know the service gets hit.  What it turns out is that if you have usedatabasenames set to true, EF itself skips using the ICandidateNamingService

https://github.com/dotnet/efcore/blob/bb583a91d7269510ecb09a4466fee61bc62280d1/src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs#L69-L77

```csharp
        _tableNamer = new CSharpUniqueNamer<DatabaseTable>(
            options.UseDatabaseNames
                ? (t => t.Name)
                : t => _candidateNamingService.GenerateCandidateIdentifier(t),
            _cSharpUtilities,
            options.NoPluralize
                ? null
                : _pluralizer.Singularize,
            caseSensitive: false);
```

So my thought would be to add something to the startup that will print a warning if you try to use both the renaming file and 
`use-database-names` at the same time since they are incompatible with eachother.

### Provide technical details

- EF Core Power Tools version: 8.1.391

- Database engine: Azure SQL

- Is T4 templates used: yes

jwyza-pi:
Resolves #2545 - Adds a warning to the output if both 'code-generation/use-database-names' is set to true and 'efpt.renaming.json' exists.

image

### DIFF
--- a/src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
+++ b/src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
@@ -71,7 +71,7 @@ internal sealed class ScaffoldHostedService : HostedService
             scaffoldOptions.ConfigFile.FullName);
         DisplayService.MarkupLine();
 
-        if (commandOptions.UseDatabaseNames && commandOptions.CustomReplacers.Count > 0)
+        if (commandOptions.UseDatabaseNames && commandOptions.CustomReplacers?.Count > 0)
         {
             configWarnings.Add("'code-generation/use-database-names' has been set to true, but a 'efpt.renaming.json' file was also found.  'use-database-names' prevents 'efpt.renaming.json' from functioning.");
         }
--- a/src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
+++ b/src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
@@ -71,6 +71,10 @@ internal sealed class ScaffoldHostedService : HostedService
             scaffoldOptions.ConfigFile.FullName);
         DisplayService.MarkupLine();
 
+        if (commandOptions.UseDatabaseNames && commandOptions.CustomReplacers.Count > 0)
+        {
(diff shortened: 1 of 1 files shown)


ErikEJ:
Thanks, great PR as always.

This should be generalized for all tools, similar to this: - so add to ReverseEngineerRunner.cs 

https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/Core/RevEng.Core.60/ReverseEngineerRunner.cs#L152

jwyza-pi:
@ErikEJ Updated to be in the ReverseEngineerRunner as requested.

ErikEJ:
@jwyza-pi Great, then it is applicable alos for VS users

Issue #1499, opened 2022-08-23
T4 customizations and tips


ErikEJ:
- [x] Enum support [sample here](https://github.com/ErikEJ/EFCorePowerTools/issues/1472#issuecomment-1220273164)
- [x]  Custom collection initializers - see https://github.com/dotnet/efcore/issues/14249 and proposal: https://github.com/ErikEJ/EFCorePowerTools/issues/1499#issuecomment-1773937177
- [x] Custom collection types: https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/managing-schemas/scaffolding/templates.md#customize-the-entity-types 
- [x] restore collection setter (N/A - fixed in EF Core)
- [x] Pure POCO classes (new POCO template added)
- [x] Multiple sets of files generated (built-in, see the wiki docs)
- [x] [sharing same template across multiple projects](https://github.com/dotnet/EntityFramework.Docs/issues/4292#issuecomment-1475204600)
- [x] [Adding Obsolete attribute to some columns](https://github.com/ErikEJ/EFCorePowerTools/issues/1750#issuecomment-1482712950)
- [x] [Fix bad mappings](https://github.com/ErikEJ/EFCorePowerTools/issues/1930)
- [x] Property renaming
- [x] [Navigation comment](https://github.com/ErikEJ/EFCorePowerTools/issues/2367)
- [ ] INotifyPropertyChanged properties
- [ ] dbcontext splitting https://github.com/dotnet/EntityFramework.Docs/pull/3995#pullrequestreview-1083179761 https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/managing-schemas/scaffolding/templates.md#entity-configuration-classes https://github.com/dotnet/EntityFramework.Docs/pull/3995 https://github.com/R4ND3LL/EntityFrameworkRuler/blob/main/src/EntityFrameworkRuler.Design/Resources/EntityTypeConfiguration.t4
- [ ] Remove default constraints

[My dotnetConf video](https://learn.microsoft.com/en-us/events/dotnetconf-2022/t4-goodness-with-entity-framework-core-7)

ErikEJ:
[Debugging templates](https://github.com/ErikEJ/EFCorePowerTools/issues/1557#issuecomment-1293514907)

sonnemaf:
I love the new T4 templates. I have used it to create C# 11 required properties. 

This is my line 104 of the `EntityType.t4`
```cs
[...]

PR #2582, opened 2024-10-16
WIP: Adds support for Using EntityTypeConfiguration files.

jwyza-pi:
Handles most of #2583. (won't be handling the GUI side of things)

TODO:

- [x] - Add Code to Ensure the scaffolded configuration files are put into the Configurations folder. 
- [x] - Pass the rootNameSpace into the scaffolding.
- [x] - Customize the DbContext.t4 to support EntityTypeConfiguration.t4
- [x] - Add a new config option: `use-t4-split` to CLI
- [x] - Warn if using both `use-t4` and `use-t4-split`.  In the case of using both, respect use-t4 instead.
- [x] - Only use the `rootNameSpace` value if `use-t4-split` is used.
- [x] - Add new zip files for .net 8 that are similar to T4_800.zip, named T4_800_Split.zip that contains the EntityTypeConfiguration.t4 and the modified DbContext.t4 (and the EntityType.t4)
  - [x] - Add the functionality to unzip this file into CodeTemplates in the event that they don't exist.

Couple items on this approach:

1. I grabbed the EntityTypeConfiguration from https://github.com/R4ND3LL/EntityFrameworkRuler/blob/main/src/EntityFrameworkRuler.Design/Resources/EntityTypeConfiguration.t4 (which was linked in https://github.com/ErikEJ/EFCorePowerTools/issues/1499.  It's for .Net7, but for whatever reason, MS didn't include it in their ef templates for .net 8 :(. 

2. You need to have context-splitting-preview turned off. 

3. ~~One note is that to make this work you need to have a modified version of DbContext.t4 that replaces the ModelLines with calling the ApplyConfiguration instead.    Not sure how we want to represent this if someone wants to use this.~~ This is now covered by having the version of the DbContext.t4 in the T4_800_Split.zip file.

~~So before:~~ (leaving the info below just for future reference as to what is different)

```csharp
     StringBuilder mainEnvironment;
    foreach (var entityType in Model.GetEntityTypes().Where(e => !e.IsSimpleManyToManyJoinEntityType()))
    {
        // Save all previously generated code, and start generating into a new temporary environment
        mainEnvironment = GenerationEnvironment;
        GenerationEnvironment = new StringBuilder();

        if (anyConfiguration)
        {
            WriteLine("");
        }

        var anyEntityTypeConfiguration = false;
#>
```

After: 

```csharp
    foreach (var entityType in Model.GetEntityTypes().Where(e => !e.IsSimpleManyToManyJoinEntityType()))
    {
        // Save all previously generated code, and start generating into a new temporary environment
        mainEnvironment = GenerationEnvironment;
        GenerationEnvironment = new StringBuilder();

        if (anyConfiguration)
        {
            WriteLine("");
        }
        
        #>
        modelBuilder.ApplyConfiguration(new Configurations.<#= entityType.Name #>Configuration());
<#
        anyConfiguration = true;
        mainEnvironment.Append(GenerationEnvironment);
        // Resume generating code into the main environment
        GenerationEnvironment = mainEnvironment;
        continue;

        var anyEntityTypeConfiguration = false;
#>

```


src/Core/RevEng.Core.60/ReverseEngineerScaffolder.cs
```diff
@@ -77,7 +78,7 @@ public SavedModelFiles GenerateDbContext(
 
                 ContextName = code.Identifier(options.ContextClassName),
                 ContextDir = outputContextDir,
-                RootNamespace = null,
+                RootNamespace = rootNameSpace,
```
jwyza-pi, line 81: This is the biggest concern I have.  From testing I did on my own project, it seemed to be OK with this, but there could be lots of scenarios that I might have missed.

src/GUI/lib/T4_800_Split/EFCore/EntityTypeConfiguration.t4
```diff
@@ -0,0 +1,270 @@
...
+<#@ import namespace="System.Linq" #>
+<#@ import namespace="System.Text" #>
+<#@ import namespace="Microsoft.EntityFrameworkCore" #>
+<#@ import namespace="Microsoft.EntityFrameworkCore.Design" #>
+<#@ import namespace="Microsoft.EntityFrameworkCore.Infrastructure" #>
+<#@ import namespace="Microsoft.EntityFrameworkCore.Scaffolding" #>
+<#@ import namespace="Microsoft.Extensions.DependencyInjection" #>
+<#@ import namespace="Microsoft.EntityFrameworkCore.Metadata.Builders" #>
+<#
+    if (!ProductInfo.GetVersion().StartsWith("7.0"))
```
ErikEJ, line 20: "8.0"

src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
```diff
@@ -84,6 +84,18 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+            DisplayService.MarkupLine("Both UseT4 and UseT4Split are set to true.  Only one of thse should be used, UseT4Split will be ignored.", Color.Yellow);
+            commandOptions.UseT4Split = false;
+        }
+
+        if (commandOptions.UseT4Split && commandOptions.UseDbContextSplitting)
+        {
+            DisplayService.MarkupLine("Both UseDbContextSplitting and UseT4Split are set to true.  Only one of thse should be used, UseT4Split will be ignored.", Color.Yellow);
+            commandOptions.UseT4Split = false;
+        }
+
```
jwyza-pi, line 98: is this where you'd want this or does it go somewhere else?

ErikEJ: Yes, this codepath is hit by both VS and the CLI, so a good place

jwyza-pi:
@ErikEJ The only thing I don't have is the unzipping of the file.  I looked and it seems like the code that drops the templates is only called via the GUI.  Unless I'm missing something.

ErikEJ:
There is a T4Helper class (I also missed that when updating to EF Core 9)

jwyza-pi:
Ok, I think it's ready now.  I did not add .Net 9 versions of the split entities largely because I don't have any of the net9 stuff setup atm.

samples/CodeTemplates/EFCore/EntityTypeConfiguration.t4
ErikEJ, line 1: Not sure about the value of adding this file here - would using it not require a new version of EntityType.t4 anyway?

samples/efcpt-schema.json
ErikEJ, line 1: Could you also add this to the new schema file (that works in VS) ?  efcpt-config.schema.json

src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
ErikEJ: Sorry, missed the location while reviewing on my mobile, these checks should live in ReverseEngineerRunner (there are some checks there already) line 191

src/GUI/RevEng.Shared/Cli/Configuration/CodeGeneration.cs
```diff
@@ -85,5 +85,9 @@ public class CodeGeneration
         [JsonPropertyOrder(200)]
         [JsonPropertyName("use-t4")]
         public bool UseT4 { get; set; }
+
+        [JsonPropertyOrder(200)]
```
ErikEJ, line 89: [JsonPropertyOrder(210)]

src/GUI/RevEng.Shared/Cli/T4Helper.cs
```diff
@@ -22,6 +22,11 @@ public static string DropT4Templates(string projectPath, CodeGenerationMode code
                 t4Version = "900";
             }
 
+            if (useEntityTypeSplitting)
+            {
+                t4Version += "_Split";
```
ErikEJ, line 27: I am fine with this potentially failing, plan to remove EF Core 6 and 7 support very soon anyway - I will create an issue to add VS and EF Core 9 support for this feature

src/GUI/lib/T4_800_Split.zip
ErikEJ, line 1: Have you verified the folder structure inside the .zip matches the other .zip files?

ErikEJ (changes requested):

ErikEJ:
This is great, just added a few comments

src/GUI/lib/T4_800_Split.zip
jwyza-pi: Yeah.  At first I had it off by 1 level, but fixed that in the most recent commit.

samples/CodeTemplates/EFCore/EntityTypeConfiguration.t4
jwyza-pi: Ah, yeah i had put that there originally before we created the 800_Split zip.  will remove.

ErikEJ (approved):
Thanks, this is a great addition

ErikEJ:
I plan to release a new non-nightly version sometime next week after .NET 9 RTW

PR #2684, opened 2024-12-02
Use Dataverse metadata to improve reverse engineering

### LINKED ISSUE #2535: Enhance Dataverse reverse engineering
improve Dataverse schema discovery by using the Dataverse client

I have tried to add 

`<PackageReference Include="Microsoft.PowerPlatform.Dataverse.Client" Version="1.1.32" />`

 to the EF Core 8 main library, only adds 2 MB of extra payload, not an issue.

@MarkMpn FYI!

MarkMpn:
When reverse engineering via the Dataverse TDS Endpoint, retrieve the metadata via the Dataverse SDK.

## Primary & foreign keys

All tables in Dataverse have a single guid primary key at the application layer, but some system tables are implemented with a compound primary key at the SQL layer which is exposed through the TDS Endpoint. This prevents the guid foreign keys in other tables being used as navigation properties.

By taking the primary key definition from the metadata instead of the SQL system views we get a more consistent view of the primary & foreign key constraints that can be reliably used to build the navigation properties as well as feeling more familiar to a Dataverse developer.

Fixes #2535

## Table & column naming

Tables and columns are exposed using their lowercase logical names in the TDS Endpoint. This leads to some hard-to-read code when we get property names like `Address1stateorprovince`.

By overwriting the names with the schema names from the metadata we get additional capitalisation so the property names tend to end up in PascalCase, e.g. `Address1StateOrProvince`.

The translation from logical to schema names is only done where there is only a difference in the casing, so the generated SQL will still be valid.

## Authentication

The TDS Endpoint and SDK use the same tokens for authentication. To avoid additional login prompts during the reverse engineering process when we switch from using a `SqlConnection` to a `ServiceClient`, this change includes an initial authentication attempt via the `ServiceClient` from which we can extract the access token. This access token can then be reused for subsequent authentication attempts via both the `SqlConnection` and `ServiceClient` without additional login prompts.


ErikEJ:
@MarkMpn Great! I wonder if it would be possible to move the Xrm dependent code to a seperate class for even lower impact?

ErikEJ:
@MarkMpn Any plans to follow up on this?

ErikEJ:
Amazing! I plan a new release to marketplace this weekend.

PR #2726, opened 2024-12-13
Fixes #2724 - Entity file getting moved incorrectly when Configuration is in the name.

### LINKED ISSUE #2724: Bug: Using T4 Context Splitting when the entity has "Configuration" at the end of the name, Incorrectly moves file.
### Provide steps to reproduce a bug

1. Set the config to use `use-t4-split`.
2. Have an Table in your list of tables name, for example, "SoftwareConfiguration".  (Any table that ends in Configuration will do though).
3. Run the scaffold.

Result:

The entity will end up getting put into a subfolder instead of the parent folder.  Eg:

output-path = "Database/EFModels"

Expected is `Database/EFModels/SoftwareConfiguration.cs`, but instead is `Database/EFModels/Configurations/SoftwareConfiguration.cs`.  

The actual IEntityTypeConfiguration file is correctly created under `Database/Configurations/SoftwareConfigurationConfiguration.cs`.  It's just the model that is wrongly placed 1 level lower than it should have.

Edit: This is likely due to the code that moves all the Configuration files post-scaffold into the Configuration folder.  The check I had written into it  looks for files that EndsWith("Configuration.cs"), which obviously breaks down in this scenario.

### Provide technical details

- EF Core Power Tools version: CLI 8.1.624

- 
- Database engine: Azure SQL

- Is T4 templates used: yes (split)

jwyza-pi:
This will resolve issue #2724.

### DIFF
--- a/src/Core/RevEng.Core.80/ReverseEngineerRunner.cs
+++ b/src/Core/RevEng.Core.80/ReverseEngineerRunner.cs
@@ -338,6 +338,26 @@ namespace RevEng.Core
             var movedFiles = new List<string>();
             foreach (var configurationFile in configurationFiles)
             {
+                // This handles the scenario where an Entity ends with the word "Configuration" such as "SoftwareConfiguration".  In this scenario
+                // there would be 2 files "SoftwareConfiguration.cs" and "SoftwareConfigurationConfiguration.cs".  The former is the entity model itself
+                // which should not be moved.  We're also checking to make sure the line isn't part of a comment block. (to handle a scenario where someone
+                // has customized the EntityType.t4 to include a comment block that points to the EntityTypeConfiguration file)
+                bool isConfigFile = false;
+                foreach (var line in File.ReadLines(configurationFile))
+                {
+                    if (line.Contains("EntityTypeConfiguration<", StringComparison.Ordinal)
+                        && !(line.Trim().StartsWith('/') || line.Trim().StartsWith('*')))
+                    {
+                        isConfigFile = true;
+                        break;
+                    }
+                }
+
+                if (!isConfigFile)
+                {
+                    continue;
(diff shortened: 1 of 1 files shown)


src/Core/RevEng.Core.80/ReverseEngineerRunner.cs
```diff
@@ -338,6 +338,26 @@ private static List<string> MoveConfigurationFiles(IList<string> files)
             var movedFiles = new List<string>();
             foreach (var configurationFile in configurationFiles)
             {
+                // This handles the scenario where an Entity ends with the word "Configuration" such as "SoftwareConfiguration".  In this scenario
+                // there would be 2 files "SoftwareConfiguration.cs" and "SoftwareConfigurationConfiguration.cs".  The former is the entity model itself
+                // which should not be moved.  We're also checking to make sure the line isn't part of a comment block. (to handle a scenario where someone
+                // has customized the EntityType.t4 to include a comment block that points to the EntityTypeConfiguration file)
+                bool isConfigFile = false;
+                foreach (var line in File.ReadLines(configurationFile))
```
ErikEJ, line 346: Would it be possible to only do this check based on the end word of the file?

ErikEJ (approved):

src/Core/RevEng.Core.80/ReverseEngineerRunner.cs
jwyza-pi: How so?  We're already only doing this check if it ends with Configuration.cs.  

Or do you mean something else that my 7AM brain missed? :)

ErikEJ: Sorry! My mistake...

PR #2767, opened 2025-01-06
Adds new option to code-generation to allow for setting the classes that are generated for Stored Procedures and Functions as internal.

### LINKED ISSUE #2741: Class modifiers on non-T4/Handlebars items (specifically sprocs)
I know in [previous](https://github.com/ErikEJ/EFCorePowerTools/issues/2450) issue requests, that folks have requested the ability to change the class access modifiers for the stored procedures.  The answer currently stands as making that a post-scaffolding operation.  Now that we have complete control over the `EntityTypeConfigurations` through the T4 templates, this is one of the last remaining places that we can't customize.

Would you be open to a enhancement PR that would add (at the very least) a setting for the class access modifiers for the related classes generated for stored procedures as an option to the CLI?  I wouldn't be able to easily do the GUI side of it, but could handle getting all the code in place so that it works for the CLI (which if I understand how the GUI works, it's mostly just hooking the GUI itself up as both share the same underlying code to do the actual generation).

If you are open to it, I'll see what it would take to accomplish, if not, that's OK too.  I understand that it adds overhead to the application that you have to maintain going forward.

(Honestly, long term I'd be great to have T4 & handlebars templates for everything, but that's a substantially heavier lift).

### Provide technical details

- EF Core Power Tools version: 8.1.624+2406c890c03798082722e1f757e32be8f82d5bb4

- Database engine: Azure SQL

- EF Core version in use: EFC8

- Is T4 templates used: yes

jwyza-pi:
Resolves #2741.

This PR adds a new configuration under code generation called `use-internal-access-modifiers-for-sprocs-and-functions`.  This setting will cause the generated classes for Functions and Stored Procedures to be `internal` rather than `public`.  This is for scenarios where T4 Templates are in use and consumers want to generate their `DBContext` and configurations as `internal` rather than `public` (often used in library style implementations where it's not desirable to expose the Data layer to consumers of a library).

A warning will be printed if `use-internal-access-modifiers-for-sprocs-and-functions` is set to true and `use-t4` or `use-t4-split` is not set. In this scenario, the `use-internal-access-modifiers-for-sprocs-and-functions` will be reset to false for the run in order to ensure a valid DBContext is generated.

The wiki docs will need to be updated with this information.

Eventually I could see this morphing to instead be a setting that could set the public/internal modifier for non-T4 scenarios as well, but that's way outside the scope of this work as it would require additional work in the stock DBContext generation.

### DIFF
--- a/src/GUI/RevEng.Shared/Cli/CliConfigMapper.cs
+++ b/src/GUI/RevEng.Shared/Cli/CliConfigMapper.cs
@@ -96,6 +96,7 @@ namespace RevEng.Common.Cli
                 UseDecimalDataAnnotation = config.CodeGeneration.UseDecimalDataAnnotation,
                 UsePrefixNavigationNaming = config.CodeGeneration.UsePrefixNavigationNaming,
                 UseDatabaseNamesForRoutines = config.CodeGeneration.UseDatabaseNamesForRoutines,
+                UseInternalAccessModifiersForSprocsAndFunctions = config.CodeGeneration.UseInternalAccessModifiersForSprocsAndFunctions,
 
                 // Not supported:
                 UseHandleBars = false,
@@ -184,6 +185,7 @@ namespace RevEng.Common.Cli
                 ProjectRootNamespace = names.RootNamespace,
                 UseDecimalDataAnnotationForSprocResult = config.CodeGeneration.UseDecimalDataAnnotation,
                 UsePrefixNavigationNaming = config.CodeGeneration.UsePrefixNavigationNaming,
+                UseDatabaseNamesForRoutines = config.CodeGeneration.UseDatabaseNamesForRoutines,
 
                 // HandleBars templates are not supported:
                 UseHandleBars = false,
--- a/src/GUI/RevEng.Shared/Cli/Configuration/CodeGeneration.cs
+++ b/src/GUI/RevEng.Shared/Cli/Configuration/CodeGeneration.cs
@@ -93,5 +93,9 @@ namespace RevEng.Common.Cli.Configuration
         [JsonPropertyOrder(210)]
         [JsonPropertyName("use-t4-split")]
         public bool UseT4Split { get; set; }
+
+        [JsonPropertyOrder(220)]
+        [JsonPropertyName("use-internal-access-modifiers-for-sprocs-and-functions")]
+        public bool UseInternalAccessModifiersForSprocsAndFunctions { get; set; }
     }
 }
--- a/samples/efcpt-config.schema.json
+++ b/samples/efcpt-config.schema.json
@@ -272,6 +272,11 @@
           "type": "boolean",
           "title": "Use stored procedure, stored procedure result and function names from the database",
           "default": true
+        },
+        "use-internal-access-modifiers-for-sprocs-and-functions": {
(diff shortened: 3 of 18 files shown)


src/GUI/RevEng.Shared/Cli/Configuration/CodeGeneration.cs
```diff
@@ -93,5 +93,9 @@ public class CodeGeneration
         [JsonPropertyOrder(210)]
         [JsonPropertyName("use-t4-split")]
         public bool UseT4Split { get; set; }
+
```
ErikEJ, line 96: I intend the ordering to be alphabetic here, so this should be moved up a bit

jwyza-pi: When I move it up, do you want me to shift the values of the `JsonPropertyOrder` down for the rest of them properties?

src/GUI/RevEng.Shared/Cli/CliConfigMapper.cs
```diff
@@ -185,8 +185,8 @@ public static ReverseEngineerOptions ToOptions(this CliConfig config, string pro
                 ProjectRootNamespace = names.RootNamespace,
                 UseDecimalDataAnnotationForSprocResult = config.CodeGeneration.UseDecimalDataAnnotation,
                 UsePrefixNavigationNaming = config.CodeGeneration.UsePrefixNavigationNaming,
-                UseDatabaseNamesForRoutines = config.CodeGeneration.UseDatabaseNamesForRoutines,
-
+                UseInternalAccessModifiersForSprocsAndFunctions = config.CodeGeneration.UseInternalAccessModifiersForSprocsAndFunctions,
```
ErikEJ, line 188: This looks like the UseDatabaseNamesForRoutines is now not set??

src/GUI/RevEng.Shared/Cli/Configuration/CodeGeneration.cs
ErikEJ: > shift the values of the JsonPropertyOrder down for the rest of them properties?

Yeah, that would be nice!

src/GUI/RevEng.Shared/Cli/CliConfigMapper.cs
jwyza-pi: Interestingly.  It's not set in the [master branch atm either](https://github.com/ErikEJ/EFCorePowerTools/blob/7a6c91d95ec3a4351366d02adff659c740d198c3/src/GUI/RevEng.Shared/Cli/CliConfigMapper.cs#L156)

jwyza-pi: To be clear, I think you saw this before I pushed another commit that undid one thing there, but the ToOptions class still doesn't set that property.  Seems like an oversight?   I'm not actually removing the setting of that property because it's not being set today.

ErikEJ: Yes, it is an oversight. I will log a new issue to get it set and add some testing

ErikEJ (approved):
LGTM

PR #2772, opened 2025-01-08
better perf in IsSqlDatabaseProjectAsync

SimonCropp:
better to do a in-mem string ends with before a async better perf in IsSqlDatabaseProjectAsync


ErikEJ:
Nice one!

PR #2773, opened 2025-01-08
remove NuGetPackageImportStamp

SimonCropp:
this is from a bug in an old version of nuget that added it to projects


ErikEJ:
@SimonCropp Thanks, everything helps 😄

PR #2819, opened 2025-01-31
Removed RevEngWizardHandler and updated ReverseEngineerHandler with Wizard code

BillKrat:
Removed preview reverse engineering menu, updated ReverseEngineerHandler with RevEngWizardHandler code [removed RevEngWizardHandler class].


ErikEJ:
@BillKrat I have published a new marketplace release, and will have a look at this PR soon.

ErikEJ:
Are some of the Dialogs not used any more and can be removed?

BillKrat:
Good question, you may have been able to glean from the PR noise that I originally removed the dialogs only to discover during testing that they were used by other handlers.  I restored them to complete testing and meant to come back to assess that; your question reminded me and below are the results of my research.

I removed the EFCoreModelDialog as it was the only dialog orphaned by wizard work.  Note: I do reuse the same interface for the Wiz3_EFCoreModelDialog page.

EFPT_ViewsHandlers

ErikEJ:
@BillKrat Just found a bug, that must be fixed berfore we can merge this

ErikEJ:
@BillKrat Maybe it would be simpler to just remove the legacy menu option and rename the wizard option? (Could even leave the "old" code behind for now)

BillKrat:
Good idea @ErikEJ, that would make it a lot simpler.  This way when you are ready to go officially with wizard you can do so with minor changes.  We can revisit old code when we look at wizards for other handlers?

PR #2824, opened 2025-02-03
Multiple configurations and connection changes fix

BillKrat:
I tested these updates with multiple configuration files and multiple connections.  I verified that changing connections work as expected.   Note that once a configuration is selected that it can be viewed on the first page tab, but cannot be changed [perhaps future update?]; the wizard has to be exited and different configuration selected.


ErikEJ:
Once you pick a configuration, that stays.  As you can see, I ask for this before anything else in the current implementation

PR #2832, opened 2025-02-05
Wizard error handling added and Rate link updated

BillKrat:
The connection error was a good find during your smoke test.  The issue was that when an error is encountered it is handled in the RevEngWizardHandler which throws it into a dialog and then exits.   This leaves the wizard hanging in memory causing the issue.   We can't change this logic because it is used by other processes that are non-wizard.

I had to create an error handler (reference image below clip) that will catch the exception, switch over to the status page, and post the error there - remaining in the wizard.  

I updated the Rates link and per our conversation will leave the status bar message as-is (for now) as the visual indication that the process is running.

https://github.com/user-attachments/assets/223fc875-2a43-47e8-bbd1-8b6952b4c057

ErrorHandler


BillKrat:
@ErikEJ you noted something about "rebase".  I've not familiar with that process, my assumption is that if I sync up my repo with yours that I'm good to go with any subsequent updates.    I'd be curious what this does and if it is necessary to keep things clean in your repo.  If so, and you could provide me guidance, I'll be sure to comply.

ErikEJ:
You can use rebase or merge. I usually start by pulling the latest master version into my fork (Github has a nice button to help with that). 

Then is start working, and when I am ready for a PR, I get the latest master again, create a branch of my master branch, and create and submit my PR.

This allows me to switch back to master and create other PR branches (for another set of files/changes)

PR #2847, opened 2025-02-13
Dialog on vs close fix

BillKrat:
Updated the Advanced and Schema dialogs so that they do not cause blocking issues on exiting VS.  Also tweaked the code so that the delete button for connections is enabled if connection is selected


src/GUI/EFCorePowerTools/Wizard/Wiz2_PickTablesDialog.cs
```diff
@@ -14,6 +7,15 @@
 using EFCorePowerTools.ViewModels;
 using Microsoft.VisualStudio.Shell;
 using RevEng.Common;
+#pragma warning disable SA1208 // System using directives should be placed before other using directives
```
ErikEJ, line 10: Just move the usings up, please

src/GUI/Shared/ViewModels/WizardDataViewModel.cs
```diff
@@ -15,10 +8,21 @@
 using EFCorePowerTools.Contracts.Wizard;
 using EFCorePowerTools.DAL;
 using EFCorePowerTools.Locales;
+using EFCorePowerTools.Wizard;
 using GalaSoft.MvvmLight;
 using GalaSoft.MvvmLight.Command;
+using Microsoft.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.Shell;
 using RevEng.Common;
+#pragma warning disable SA1208 // System using directives should be placed before other using directives
```
ErikEJ, line 17: Just move them up

ErikEJ (changes requested):
LGTM, but please remove the added pragmas

BillKrat:
Yea, that was an interesting one; a game of wack-a-mole.   When I clicked save they were sorted to the bottom and then I got the error.  I'll have to look at what I have turned on so that it will not change things as I save...

I'm curious if I did something wrong.  My branch master is showing 9 commits "ahead" and this PR is showing 10 commits, where I only changed 3 files and would only expect the one commit.....  I think I only looked at the pointer below without taking into consideration I was ahead.....

Should I restart, have my master first match yours, create a local branch of it, and resubmit a PR?  

image

ErikEJ:
No need to restart, just update the using statement order

ErikEJ:
What "tool" do you use? Recharger? Itcan be disabled.

src/GUI/EFCorePowerTools/Wizard/Wiz1_PickServerDatabaseDialog.cs
```diff
@@ -12,6 +8,12 @@
 using EFCorePowerTools.ViewModels;
 using Microsoft.VisualStudio.Shell;
 using RevEng.Common;
+#pragma warning disable SA1208 // System using directives should be placed before other using directives
```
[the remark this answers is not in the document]
BillKrat: Moved

src/GUI/EFCorePowerTools/Wizard/Wiz2_PickTablesDialog.cs
BillKrat: moved

src/GUI/Shared/ViewModels/WizardDataViewModel.cs
BillKrat: moved

BillKrat:
image
I use Visual Studio 2022 - seems by default it sets up the following profile (we keep this on for work)

ErikEJ (approved):

Issue #2859, opened 2025-02-16
Focus issue with Wizard


ErikEJ:
The status screen shows, but once the Power Tools readme.md is opened in the editor, the Wizard gets hidden behind VS and running the Wizard again starts a new one. 

They are all closed when exiting VS.

@BillKrat FYI

PR #2860, opened 2025-02-16
Focus issue with wizard

BillKrat:
Focus issue with Wizard #2859 - updated so wizard stays on top and only once instance can run at a time


src/GUI/Shared/Handlers/ReverseEngineer/RevEngWizardHandler.cs
```diff
@@ -114,8 +116,14 @@ public async Task ReverseEngineerCodeFirstLaunchWizardAsync(WizardEventArgs wiza
                 // which this class implements.  The wizard pages will use the BLL to process
                 // data using existing business logic; to simplify wizard refactor this handler
                 // serves as the BLL (and DAL)
-                var wizard = new WizardDialogBox(this, wizardArgs, wizardViewModel);
-                wizard.Show();
+                if (!WizardIsRunning)
+                {
+                    WizardIsRunning = true;
+                    var wizard = new WizardDialogBox(this, wizardArgs, wizardViewModel);
+                    wizard.Owner = Application.Current.MainWindow;
```
ErikEJ, line 124: Is this the solution for the focus issue?

ErikEJ:
LGTM

src/GUI/Shared/Handlers/ReverseEngineer/RevEngWizardHandler.cs
BillKrat: Yes, on line 123 that ensures it stays on top of the current application.

PR #2902, opened 2025-03-13
Reveng parameterless scalar functions to be callable directly from the DbContext

### LINKED ISSUE #2921: SQL Function support
Hi,

I apologise if my research skills are subpar, but I am struggling to see anything on this.

I am looking to use some SQL functions within my .NET API - However, it appears that the EFCore Power Tools doesn't seem to want to generate the function wrapper?

In the  `FareshareCoreDbContext.Functions.cs` 

```
        [DbFunction("Loyalty_CurrentDateTime", "dbo")]
        public static DateTime? Loyalty_CurrentDateTime()
        {
            throw new NotSupportedException("This method can only be called from Entity Framework Core queries");
        }
```

You can see that the method hasn't been generated by the power tools. The function is pretty simple and just purely gets the DateTime from the DB

```
CREATE OR ALTER   FUNCTION [dbo].[Loyalty_CurrentDateTime]
(
)
RETURNS datetime
AS
/*
Core? YES

Usage:
SELECT dbo.Loyalty_CurrentDateTime()

*/
BEGIN

DECLARE	@CurrentDateTime datetime

-- Get the current date and time --
SELECT
@CurrentDateTime = GETDATE()

RETURN (@CurrentDateTime)

END
GO
```

I know I could technically build my own wrapper but if I can avoid that, that'd be good.

Thanks,
Haydon

JanRomero:
Sorry if this is dumb, but I would like to avoid having to write…

```csharp
await Database.SqlQueryRaw<DateTime>("select Value = dbo.MyIncredibleFunction()").SingleAsync();
```

This would make it so you could instead just write:

```csharp
await dbContext.MyIncredibleFunctionAsync();
```

Works only for parameterless functions, but I imagine it could be made to work with params by duplicating the relevant code from Stored Procedures.

Cheers

fixes #2921


ErikEJ:
@JanRomero the functions are already mapped??

But perhaps not directly callable?

JanRomero:
@ErikEJ thanks for the quick response!

Yes, since TVFs can be called directly like
```csharp
var results = await dbContext.MyTableValuedFunction(myParam, myOtherParam).ToListAsync();
```
and Stored Procs work similarly, but scalar functions can’t be called. Just thought it would be nice if the API were more consistent, as the workaround using raw sql is kind of clunky.

For example, I have a scalar function that returns the latest timestamp across several tables to see if a cache needs refreshing. So it’s a function you’d just like to call on its own, as opposed to those used within queries like `dbo.IsoWeek(@date date)` or something.

I don’t really expect this puny PR to be merged since it only handles parameterless functions. Just floating the idea. I suppose I could try and add the code for params if you can see this being a thing.

ErikEJ (approved):

ErikEJ:
@JanRomero Thanks for a great contribution

PR #2934, opened 2025-04-09
fix: ask_a_question template options values

Gean-Vasconez:
GitHub does not allow for options to be boolean values, instead they should be strings.

Fixing this so I can issue a question following the expected templating.


ErikEJ:
Thanks!

Issue #2931, opened 2025-04-08
Reverse Engineer Wizard cause Visual Studio 2022 restart


doggy8088:
### Bug description / exact reproduction steps

I read from "Config file", then hit `Next`, then Visual Studio 2022 get hang, then restart.

![Image](https://github.com/user-attachments/assets/3e39a10e-2b24-4ed2-9b7d-eb703b59f1e5)

Sorry, I don't know how to gather logs.

### Your code

```csharp
N/A
```

### Stack traces

```text

```

### EF Core Power Tools version

2.6.961

### EF Core Power Tools CLI version

N/A

### Database provider

Microsoft.EntityFrameworkCore.SqlServer

### Target framework

.NET 8.0

### IDE

Visual Studio 2022 v17.13.5

### Additional features in use

- [ ] T4 templates
- [ ] SQL Server .dacpac
- [ ] Handlebars template

ErikEJ:
Could you show a screen shot of "read config file"??

ErikEJ:
@BillKrat Ping?

BillKrat:
Do we have confirmation?  Or are you wanting me to triage and attempt to duplicate @ErikEJ

ErikEJ:
@BillKrat Lets await details/confirmation!

@doggy8088 Ping?

doggy8088:
@ErikEJ I recorded a video that can reproduce this issue every time, please take a look: https://youtu.be/y-jovDcq0iw

BillKrat:
Perfect, @ErikEJ, I’ll triage and resolve.

BillKrat:
Thank you @doggy8088 for your patience, time, and help in isolating this.   This issue is contributed to my poor design and by failing to make the process comprehensive; your video was instrumental in helping me discover this...

For all practical purposes that "Config file" tab is informational only.   The real process is to "Add" a connection [see add button next to pointer below] and then select the connection that you added to the dropdown so that it is visible.   I should also not have allowed the [next] button to be enabled until there was a selected connection.   I can definately understand the confusion...

Please add a connection if you have not already done so, then select the connection so that it is visible on the "Connection String" tab pointed to below.   Once that is set then please click the Next button and see if you get expected results.

I apologize for any lost time that this design issue caused you.
BillKrat

![Image](https://github.com/user-attachments/assets/2f709a98-5e79-469d-8116-f72e6e905228)

BillKrat:
[...]

PR #2946, opened 2025-04-29
#2931 Next disabled until connect select

BillKrat:
Added new BoolNullConverter so that it can control the Next Button IsEnabled value (only enabled if SelectedConnectionString is not null).  On the Config tab I made it more apparent that only the configuration path is displayed - removed combo and replaced with textbox.

### DIFF
--- a/src/GUI/EFCorePowerTools/Wizard/Wiz1_PickServerDatabaseDialog.xaml
+++ b/src/GUI/EFCorePowerTools/Wizard/Wiz1_PickServerDatabaseDialog.xaml
@@ -183,21 +183,47 @@
                         <RowDefinition Height="10" />
                     </Grid.RowDefinitions>
 
-                    <Label Grid.Column="1" Grid.Row="1">Config File:</Label>
+                    <Label Grid.Column="1"
+                           FontWeight="bold"
+                           Grid.Row="1">Selected Configuration File:</Label>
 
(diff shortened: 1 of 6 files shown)


src/GUI/EFCorePowerTools/Wizard/Wiz1_PickServerDatabaseDialog.xaml
```diff
@@ -182,21 +183,47 @@
                         <RowDefinition Height="10" />
                     </Grid.RowDefinitions>
 
-                    <Label Grid.Column="1" Grid.Row="1">Config File:</Label>
+                    <Label Grid.Column="1"
+                           FontWeight="bold"
+                           Grid.Row="1">Selected Configuration File:</Label>
 
-                    <ComboBox Grid.Column="2" Grid.Row="1"
+                    <!--<ComboBox Grid.Column="2" Grid.Row="1"
```
ErikEJ, line 190: Remove commented out code?

BillKrat: I could, was wondering if we didn't want to have the ability at some point to change configuration "without" having to exit the app.   Would be a medium effort.   Your call, with everyone gone Diane and I are going on a date night, I can do it in a.m. if you would like it removed.

Issue #3122, opened 2025-09-15
Wizard: Customize code using templates option not properly persisted


RickBogard:
### Bug description / exact reproduction steps

Selecting the T4 Template in the legacy UI. I deselect the "Customize code using templates" option, yet EFPT still uses the T4 Template because I am getting the warning that the Template Version 8.0 is not found. It also still has the "Customize code using templates" still selected the when I launch the tool using the Reverse Engineer menu option a second time.

This occurs using ver 2.6.1130.

### Your code

```csharp
This is the part of the config file where the UseT4Split is stored.  It does not change when I deselect it in the Wizard UI.

   "UiHint": null,
   "UncountableWords": null,
   "UseAsyncStoredProcedureCalls": true,
   "UseBoolPropertiesWithoutDefaultSql": false,
   "UseDatabaseNames": true,
   "UseDatabaseNamesForRoutines": true,
   "UseDateOnlyTimeOnly": false,
   "UseDbContextSplitting": false,
   "UseDecimalDataAnnotationForSprocResult": true,
   "UseFluentApiOnly": false,
   "UseHandleBars": false,
   "UseHierarchyId": false,
   "UseInflector": true,
   "UseInternalAccessModifiersForSprocsAndFunctions": false,
   "UseLegacyPluralizer": false,
   "UseManyToManyEntity": true,
   "UseNoDefaultConstructor": false,
   "UseNoNavigations": false,
   "UseNoObjectFilter": false,
   "UseNodaTime": false,
   "UseNullableReferences": true,
   "UsePrefixNavigationNaming": false,
   "UseSchemaFolders": true,
   "UseSchemaNamespaces": false,
   "UseSpatial": false,
   "UseT4": false,
   "UseT4Split": true
```

### Stack traces

```text

```

### EF Core Power Tools version

2.6.1130

### Target framework

.net 8.0

### IDE

VS IDE 17.14.14

### Additional features in use

- [x] T4 templates
- [ ] SQL Server .dacpac
- [ ] Handlebars template

ErikEJ:
Have you tried renaming or removing the CodeTemplates folder?

RickBogard:
Yes, this occurs even after doing that.

ErikEJ:
Please share exact repro steps or a repro project

ErikEJ:
@RickBogard Ping?

RickBogard:
[...]

PR #3157, opened 2025-09-27
#3122 Added missing logic to handle "Customize code using templates" …

BillKrat:
…checkbox and combo.   TemplateTypeList combo no longer adds duplicate entries when navigating prev/next.

Note: tested to ensure all checkbox and input fields are persisted.


ErikEJ (approved):

ErikEJ:
Thanks!

Issue #3169, opened 2025-10-03
General Humanizer vocab configuration


Phen-Ro:
### What problem are you trying to solve?

I'm having trouble fixing the inflection of tables that end with the word "locus", the plural of which is "loci".  I'm getting unfortunate entity names like "SuchAndSuchLocu", for example.  This is part of a database with many other tables given singular names, so simply turning off all inflections isn't a great option.  The legacy pluralizer isn't any better.

I've tried the efpt.renaming.json configuration, but `{ "Name": "SuchAndSuchLocus", "NewName": "SuchAndSuchLocus" }` doesn't have an effect, and in any case I have to way to tell EFCPT/Humanizer that a SuchAndSuchLocus collection should be named SuchAndSuchLoci.

### Describe the solution you'd like

What I'd really like is to open up Humanizer configuration more. 

I see in the [ServiceProviderBuilder](https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/Core/RevEng.Core.80/ServiceProviderBuilder.cs#L102) that the `uncountable-words` configuration from the config json file translates to the `AddUncountable()` method to the Humanizer dictionary.

I'd also like access to the `AddIrregular()`, `AddPlural()`, and `AddSingular()` methods, especially `AddIrregular()`.

If those methods were exposed in the config file in a similar way as UncountableWords, I think it would solve all of the numerous pluralization issues I've waded through in the GitHub issues here.

ErikEJ:
@Phen-Ro you could start by adding a test that proves that your proposal would fix your concrete issue. I am reluctant to increase the codebase just because we can.

Phen-Ro:
I had a heck of a time spinning up the solution, between bad advice from Microsoft on how to use .Net 10 in Visual Studio, some apparently missing package references for Microsoft.VisualStudio.TemplateWizardInterface, and a few unit tests that failed (GetRelativePath1, GetRelativePath2, and LoadedCommand_Executed_OnlyDatabaseConnections).  I also couldn't make heads or tails on how to use the ScaffoldingTester solution.  If the adding to the ScaffoldingTester is what you mean for me to do, I may need some hand holding.

Nevertheless, I added `irregular-words`, `plural-rules`, and `singular-rules` to the `replacements` config and hooked them up in the same manner as `uncountable-words`.  (In the code, I added IrregularWords before UncountableWords, and PluralRules and SingularRules after UncountableWords because that is the order that Humanizer uses in its code.)
[...]

PR #3170, opened 2025-10-05
General Humanizer vocab configuration

Phen-Ro:
Add `irregular-words`, `plural-rules`, and `singular-rules` to the `replacements` config to expose Humanizer's AddIrregular(), AddPlural(), and AddSingular().

See #3169.


ErikEJ:
@Phen-Ro Why is it a draft?

src/GUI/RevEng.Shared/ReverseEngineerCommandOptions.cs
```diff
@@ -59,8 +59,14 @@ public class ReverseEngineerCommandOptions
 
         public bool UseLegacyPluralizer { get; set; }
 
+        public List<(string Singular, string Plural, bool MatchEnding)> IrregularWords { get; set; }
```
ErikEJ, line 62: Why not use the class here?

src/GUI/RevEng.Shared/ReverseEngineerCommandOptions.cs
```diff
@@ -59,8 +59,14 @@ public class ReverseEngineerCommandOptions
...
+
         public List<string> UncountableWords { get; set; }
 
+        public List<(string Rule, string Replacement)> PluralRules { get; set; }
```
ErikEJ, line 66: Why not classes?

ErikEJ:
@Phen-Ro Please add tests to CustomHumanizerPluralizerTest.cs to prove that this actually fixes your concrete issue, thanks.

Phen-Ro:
I marked it as a draft because I don't use GitHub that much and the potential for me pressing bad buttons is pretty high lol

I've added some tests to CustomHumanizerPluralizerTest.cs.

Every time Visual Studio opens `efcpt-config.schema.json`, it reformats the spaces.  Do you care if that comes along for the ride?  I can re-amend that commit otherwise.

For the tuples vs classes, I just used tuples because the usage of the command options is pretty far away from the json serialization concerns, so it seemed weird to use the same class, and I didn't feel like declaring another one.  But I don't mind changing that if you want.

ErikEJ:
@Phen-Ro 

> Do you care if that comes along for the ride?

No problem

> he usage of the command options is pretty far away from the json serialization concerns

The class with the tuples is actually used fro serilization from the VS extension - but I will take care of that in a future PR

ErikEJ (approved):

ErikEJ:
@Phen-Ro Thanks a lot for this

darrenhollick:
@Phen-Ro or @ErikEJ I am having a similar problem where I have a table named "Faxes" and it is not properly singularizing the entity to "Fax".  I tried looking through your changes which have now been merged and I am hoping you could tell me how I can modify my `efpt.config.json` file to fix the issue.

I tried adding this to my file but that did not seem to have any effect:
```json
{
  ...
  "IrregularWords": [
     {
       "MatchEnding": false,
       "Plural": "Faxes",
       "Singular": "Fax"
     }
  ],
  ...
}
```

To be clear I'm looking for the collection on the context to be named "Faxes" and the entity class to be named "Fax".  In a perfect world this would also correct any related property names as well. (e.g. a property on a table with a relationship to Faxes such as a Users table will generate a User entity with a property of "Faxes")

Can you offer me any guideance?  Thank you in advance!

ErikEJ:
@darrenhollick This does not relate to renaming of navigation properties

Phen-Ro:
Iirc, the naming of the navigation property collections should still be influenced.  @darrenhollick, is that configuration block inside a `replacements` object?  Without trying this myself, I would expect something like this to work:
```json
{
  "$schema": "https://raw.githubusercontent.com/ErikEJ/EFCorePowerTools/master/samples/efcpt-config.schema.json",
  "code-generation": {
    ...
  },
  "replacements": {
    "irregular-words": [
        { "singular": "Fax", "plural": "Faxes" }
    ]
  },
  ...
}
```

nathanhqws:
Sorry if this is not the right place but given the recent posts I thought I would tap into it.

I am having a similar issue where I am just trying to stop pluralization changes of "Data" into "Datum".

I added

```
"UncountableWords": [
      "Status",
      "Data"
   ],
```
   
   To my `efpt.config.json` but it made no difference, the single object instance has Datum at the end. How can I use the replacements section, does that just go in `efpt.config.json` as it does not seem to match the casing nor is there an existing section in there?
   
  All I want is "Data" to be used for both the singular and plural.

ErikEJ:
@nathanhqws if you have issue and need free help, please create a new issue with actionable reproduction details.

ErikEJ:
@nathanhqws if you have issue and need free help, please create a new issue with actionable reproduction details.

PR #3343, opened 2026-04-01
chore: move sln to slnx

nmummau:
Not really sure how you build/release your tool but figured you would welcome the new slnx format.
I'm just switching to the slnx format, which was done automatically with `dotnet sln migrate`


ErikEJ:
@nmummau please have a look at the build error,  maybe some of the pipelines need an update?

PR #3346, opened 2026-04-01
chore: remove unused usings

nmummau:
Only touching test files. I used `dotnet format style`, a little trick to remove unused usings.
I did have to copy the /src/.editorconfig to /test/.editorconfig temporarily to get this to work.

Possibly .editorconfig should be moved to the repo root


ErikEJ:
These generated files are not subject to formatting and ignored by analyzers

Issue #3361, opened 2026-04-03
Live Database versus Dacpac?


nmummau:
### Question

When comparing the models that get generated when reverse engineering from a live mssql database versus a Dacpac, I'm noticing many differences. Is this expected? It might be tricky to list all of them but I'll try.

### 1. Extended Properties with apostrophe
> A fix is in progress within https://github.com/ErikEJ/EFCorePowerTools/pull/3364
```sql
EXEC sp_addextendedproperty
    @name = N'MS_Description',
    @value = N'The location''s address',
    @level0type = N'SCHEMA', @level0name = N'dbo',
    @level1type = N'TABLE',  @level1name = N'LocationAddress',
    @level2type = N'COLUMN', @level2name = N'SomeColumn';
GO
```
The Live DB turns this into 
```csharp
    /// <summary>
    /// N'The location&apos;s address
    /// </summary>
    public byte N'SomeColumn { get; set; }
```
Dacpac uses duplicate `&apos;&apos;`
```csharp
    /// <summary>
    /// N'The location&apos;&apos;s address
    /// </summary>
    public byte N'SomeColumn { get; set; }
```

### 2. Column with a default constraint
```sql
PreviousEventPosition  BIGINT NOT NULL CONSTRAINT DF_PreviousEventPosition DEFAULT (0),
```
Live DB
```csharp
public long PreviousEventPosition { get; set; } = 0L;
```
dacpac
```csharp
public long PreviousEventPosition { get; set; };
```

### 3. Stored procedure data type

Live DB
```csharp
Task<List<StoGetStatusMetricCountsResult>?> StoGetStatusMetricCountsAsync(OutputParameter<int>? returnValue = null, CancellationToken? cancellationToken = default);
```
Dacpac
```csharp
Task<int> StoGetStatusMetricCountsAsync(OutputParameter<int>? returnValue = null, CancellationToken? cancellationToken = default);
```

### 4. Stored procedure execute

Live DB uses `_context.SqlQueryAsync<T>`
```csharp
var _ = await _context.SqlQueryAsync<StoGetStatusMetricCountsResult>("EXEC @returnValue = [base].[StoGetStatusMetricCounts]", sqlParameters, cancellationToken);
```

Dacpac uses `_context.Database.ExecuteSqlRawAsync`
```csharp
[...]

PR #3367, opened 2026-04-05
fix(dacpac): preserve multiple table triggers

nmummau:
Fixes part of #3361

  Capture all table triggers when reverse engineering from dacpac files
  instead of only the first trigger reference.

  Adds a fixture-backed regression test covering a table with multiple
  triggers in the AdventureWorks2014 dacpac.


ErikEJ:
LGTM, just need to fix the merge conflict

nmummau:
> LGTM, just need to fix the merge conflict

@ErikEJ Fixed conflicts.

PR #3373, opened 2026-04-06
chore: remove unused deps from EfPlayground

nmummau:
The EfPlayground project still builds without these deps so I'm pretty sure we can remove them to save on dep maintenance in the future.


ErikEJ:
@nmummau LGTM  but there are merge conflicts now

Issue #3376, opened 2026-04-06
Dacpac scaffolding strips column default values that live DB scaffolding preserves


nmummau:
### Description

  SqlServerDacpacDatabaseModelFactory filters out column defaults that match CLR type defaults via FilterClrDefaults(), but PatchedSqlServerDatabaseModelFactory (live DB path) does not. This causes the two scaffolding sources to produce different DatabaseColumn.DefaultValue and DatabaseColumn.DefaultValueSql values for the same schema.

### Steps to reproduce

  Given a table with a non-nullable numeric column that has a default constraint:

```sql
CREATE TABLE readModel.ItemInventoryNodeBalance (
    PreviousEventPosition BIGINT NOT NULL
    CONSTRAINT DF_readModel_ItemInventoryNodeBalance_PreviousEventPosition DEFAULT (0)
);
```

  Scaffold from a live
  Scaffold from a dacpac 

#### Expected behavior

  Both paths produce identical DatabaseColumn values:
  - DefaultValue: 0L (parsed CLR object)
  - DefaultValueSql: "((0))" or "0" (SQL expression)

#### Actual behavior

| Property | Live DB | Dacpac | 
| -- | -- | -- |
| DefaultValue | 0L | null |
| DefaultValueSql | "((0))" | null |

  This affects all non-nullable columns whose defaults match CLR defaults:
- 0 for numeric types
- false/0 for bit
- '0001-01-01' for date
- '00000000-0000-0000-0000-000000000000' for uniqueidentifier
- etc.

### Root cause
  https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/Nupkg/ErikEJ.EntityFrameworkCore.8.SqlServer.Dacpac/Scaffolding/SqlServerDacpacDatabaseModelFactory.cs#L414-L472 in the dacpac path explicitly returns null for defaults that match CLR type defaults. The live DB path has no equivalent filtering — it passes the raw sys.default_constraints.definition directly to DefaultValueSql and through TryParseClrDefault() to DefaultValue.

  The filtering happens at line 713:
  var defaultValue = def != null ? FilterClrDefaults(systemTypeName, isNullable, def.Expression) : null;

  ### Suggested fix

  Replace FilterClrDefaults with StripParentheses so the dacpac path passes defaults through unfiltered, matching the live DB behavior:

  var defaultValue = def != null ? StripParentheses(def.Expression) : null;

[...]

PR #3377, opened 2026-04-06
Dacpac path passes defaults

nmummau:
Propose a fix for #3376

### Fix
Replace FilterClrDefaults with StripParentheses so the dacpac path passes defaults through to TryParseClrDefault the same way the live DB path does.


ErikEJ:
@nmummau do we have tests covering this all ready?

nmummau:
@ErikEJ updated an existing test that was sort of hiding this issue, and I added a new test.

Issue #3380, opened 2026-04-07
FluentAssertions went commercial in v8


nmummau:
### What problem are you trying to solve?

Since FluentAssertions went commercial in v8, I think that should be motivation to move onto something else for assertions.

### Describe the solution you'd like

Can we migrate all FluentAssertions assertions to to Nunit 4 assertions? Seems reasonable for all tests to use the same assertion library.

ErikEJ:
@nmummau I would actually prefer migrating to XUnit V3 assertions

Issue #3379, opened 2026-04-07
Migrate to NUnit4 in NUnitTestCore.csproj


nmummau:
### What problem are you trying to solve?

I noticed that the project `NUnitTestCore.csproj` still uses some old dependencies. Most of them can be updated in place but AutoFixture.NUnit3 requires a migration to use AutoFixture.NUnit4.

This issue is to track that work

### Describe the solution you'd like

use the migration guide to migrate to v4
https://docs.nunit.org/articles/nunit/release-notes/Nunit4.0-MigrationGuide.html

ErikEJ:
@nmummau Much prefer migarion to XUnit V3

nmummau:
> @nmummau Much prefer migarion to XUnit V3 

What would you think about merging my NUnit 4 changes in, and getting Copilot in GitHub to migrate to xunit.v3?

ErikEJ:
@nmummau I would prefer to get the current projects migrated directly? Would that not be similar?

nmummau:
> @nmummau I would prefer to get the current projects migrated directly? Would that not be similar?

Currently there is a mixture of assertion frameworks: NUnit 3, NUnit 4, FluentAssertions.

I'm sure Copilot could port all of those to xunit.v3 but the work I did already aligned everything to NUnit 4, so Copilot might have an easier time doing that single port.

Open to whatever though.

ErikEJ:
@nmummau  Fair point, if you commit to a follow up PR?

nmummau:
> @nmummau  Fair point, if you commit to a follow up PR?

Sure I can do that. Is copilot on GitHub an option? Like can we have it fix it for us and create a PR? I've seen some other Microsoft OSS projects doing that but maybe that's not available for free to all of OSS?

ErikEJ:
@nmummau I will ask Copilot for help with this!

nmummau:
@ErikEJ how do you kick that off? Only maintainers I'd assume?

ErikEJ:
@nmummau Correct, I have some credits via my company that I can use, so only me as a maintainer can kick it off

nmummau:
@ErikEJ if you can, point Copilot at this guide https://xunit.net/docs/getting-started/v3/migration

PR #3381, opened 2026-04-07
test: migrate core tests to NUnit 4 and remove FluentAssertions

nmummau:
test: migrate core tests to NUnit 4
  - upgrade NUnitTestCore to NUnit 4 with AutoFixture.NUnit4
  - convert core tests to modern Assert.That syntax

  This is for issue #3379

test: remove FluentAssertions
  - convert core tests to Assert.That syntax
  - remove FluentAssertions from NUnitTestCore

  This fixes issue #3380


ErikEJ:
I would actually prefer to migrate to XUnit V3 and just simple XUnit asserts TBH

nmummau:
> I would actually prefer to migrate to XUnit V3 and just simple XUnit asserts TBH

Let's do that then. I only suggested NUnit v4 cause that's what is already in here. But I'm more familiar with xunit.v3 anyway.

PR #3384, opened 2026-04-08
fix(sqlserver): align live-db stored procedure scaffolding with dacpac for temp tables and parameter sizing

nmummau:
This PR fixes two SQL Server stored procedure scaffolding issues in the live-database path where behavior currently differs from DACPAC reverse engineering.

  ## Problem

  ### 1. Temp-table-backed stored procedures scaffold correctly from DACPAC but fail from live-db

  For stored procedures that use #temp tables, the DACPAC path can successfully infer the result-set shape, while the live-db path can fail result-set discovery and generate an empty result model.

  Minimal repro:
```sql
  CREATE PROCEDURE dbo.StoGetSomeData
  AS
  BEGIN
      SET NOCOUNT ON;

      CREATE TABLE #Temp
      (
          SomeName NVARCHAR(100),
          SomeValue NVARCHAR(100),
      );

      SELECT
          t.SomeName,
          t.SomeValue
      FROM #Temp t;

      DROP TABLE #Temp;
  END;
  GO
```

  Current DACPAC behavior:

  - result-set discovery succeeds
  - generated model contains SomeName and SomeValue

  Current live-db behavior:

  - result-set discovery fails with:
    warning: Unable to get result set shape for procedure 'dbo.StoGetSomeData'. Invalid object name '#Temp'.
  - generated result type is empty

  Example of the incorrect live-db output:

```csharp
  public partial class StoGetSomeDataResult
  {
  }
```
  ### 2. Unicode parameter sizes are incorrect in the live-db path

  For parameters such as @SearchTerm NVARCHAR(50) = NULL, the DACPAC path scaffolds Size = 50, but the live-db path scaffolds Size = 100.

  Minimal repro:

```sql
  CREATE PROCEDURE dbo.StoGetSomeDataWithParameters
      @CategoryId INT,
      @SearchTerm NVARCHAR(50) = NULL,
      @Amount DECIMAL(10,2) = 0
  AS
  BEGIN
      SET NOCOUNT ON;

      CREATE TABLE #Temp
      (
          CategoryId INT NOT NULL,
          SearchTerm NVARCHAR(50) NULL,
          Amount DECIMAL(10,2) NOT NULL
      );

      INSERT INTO #Temp
      (
          CategoryId,
          SearchTerm,
          Amount
      )
      VALUES
      (
          @CategoryId,
          @SearchTerm,
          @Amount
      );

      SELECT
          t.CategoryId,
          t.SearchTerm,
          t.Amount
      FROM #Temp t;

      DROP TABLE #Temp;
  END;
  GO
```

  Expected / DACPAC behavior:

```csharp
  new SqlParameter
  {
      ParameterName = "SearchTerm",
      Size = 50,
      Value = searchTerm ?? Convert.DBNull,
      SqlDbType = System.Data.SqlDbType.NVarChar,
  }
```

  Current live-db behavior:

```csharp
  new SqlParameter
  {
      ParameterName = "SearchTerm",
      Size = 100,
      Value = searchTerm ?? Convert.DBNull,
      SqlDbType = System.Data.SqlDbType.NVarChar,
  }
```

  Root cause:

  - live-db reads parameter length from sys.parameters.max_length
  - SQL Server reports Unicode lengths in bytes, not characters
  - nvarchar(50) therefore comes back as 100
  - that raw value was being passed through without normalization

  ## Fix

  This PR updates the live-db stored procedure scaffolding path to behave more like the DACPAC path.

Changes:

  - add fallback result-set discovery for live-db stored procedures when metadata discovery fails
  - keep the existing metadata discovery behavior as the primary path:
    - use `sp_describe_first_result_set` when alternate stored procedure discovery is enabled
    - otherwise use the normal schema-based live-db metadata discovery path
- when metadata discovery fails in known temp-table-related cases, fall back to parsing the stored procedure definition from `sys.sql_modules`
  - add a shared stored-procedure result-set factory so DACPAC and live-db use the same temp-table-aware parsing logic
  - normalize Unicode parameter lengths in the live-db path so `nchar`, `nvarchar`, and `sysname` use character counts instead of byte counts

  ## Result

  After this change:

  - temp-table-backed stored procedures can scaffold result sets from live-db instead of generating empty result models
  - when metadata discovery fails in known temp-table-related cases, live-db can recover by parsing the stored procedure definition
  - live-db and DACPAC scaffolding produce consistent Unicode parameter sizes
  - DACPAC and live-db stored procedure inference share more of the same parsing logic

  ## Testing

  Added regression coverage for:

  - temp-table-backed stored procedure result-set discovery
  - parameterized procedures
  - multiple result sets
  - Unicode parameter length normalization for live-db metadata

  Also added a small Docker/playground repro project to validate both paths against the same procedures.

  ## Notes

  Most of the diff size is supporting test/repro material, not runtime code:

  - product code: 5 files
  - tests / playground / generated output: 17 files

  This fixes the live-db/DACPAC inconsistency reported in #3384, and also addresses the Unicode parameter sizing issue discovered while investigating #3385.


ErikEJ:
@nmummau How does this interact with the current possibility to set discovery type per stored procedure (as documented) ?

src/Core/RevEng.Core.80/Routines/Procedures/SqlServerStoredProcedureModelFactory.cs
```diff
@@ -83,8 +243,31 @@ private static List<List<ModuleResultElement>> GetAllResultSets(SqlConnection co
 
                 sqlCommand.Parameters.Add(param);
             }
+        }
 
-            using var schemaReader = sqlCommand.ExecuteReader(CommandBehavior.SchemaOnly);
+        /// <summary>
```
ErikEJ: Why not just use CommandBehavior.SchemaOnly ?

ErikEJ (commented):
LGTM, just a couple of questions

ErikEJ:
@nmummau Will this change the wording in the wiki?

nmummau:
> @nmummau How does this interact with the current possibility to set discovery type per stored procedure (as documented) ?

Good question. I didn't test that. Let me test that today and let you know

nmummau:
> @nmummau Will this change the wording in the wiki?

It probably will. Let me look over that and make a suggested change to the wiki. I should be able to get to that today.

src/Core/RevEng.Core.80/Routines/Procedures/SqlServerStoredProcedureModelFactory.cs
nmummau: `CommandBehavior.SchemaOnly` is already the first thing this code tries in GetAllResultSets(). The problem is that for some procedures, especially ones that create and then select from #temp tables, SQL Server still fails metadata discovery on that path and throws `Invalid object name '#Temp'` instead of returning schema metadata.

My thought process here is to use this fallback only for the cases where `SchemaOnly` did not work.

My intent is not to replace `CommandBehavior.SchemaOnly`, but to preserve it as the primary path and add a fallback when SQL Server cannot describe the result set correctly

More specifically:

  - `SchemaOnly` works by asking SQL Server for result metadata from the stored procedure call.
  - For temp-table-backed procedures, SQL Server can fail to describe the shape during that process, even though the DACPAC path can infer it from the definition.
  - `FMTONLY` gives SQL Server a different metadata-discovery path, and if that still fails, the final fallback is parsing the procedure definition from `sys.sql_modules`.

ErikEJ: But schemaonly is the same a FMTONLY!

https://learn.microsoft.com/en-us/dotnet/api/system.data.commandbehavior?view=net-10.0

ErikEJ: There are two known ways to discover the result shape:

1: sp_describe_first_result_set

2: SET FMTONLY ON

and you are now adding parsing of procedure defintion from sys.modules in the live database

ErikEJ: Based on years of experience, SET FMTONLY works best in most cases, so sp_describe_first_result_set is a fallback

nmummau: > But schemaonly is the same a FMTONLY!

Oh! That simplifies it I suppose. Let me test.

nmummau: > Based on years of experience, SET FMTONLY works best in most cases, so sp_describe_first_result_set is a fallback

@ErikEJ  do you mean "sys.sql_modules" is the fallback?

ErikEJ: No, usage of sys.sql_modules is the new feature that you are adding!

nmummau: @ErikEJ thanks for helping me find that redundancy. It's been patched now and all tests still pass which proves that sql_modules is still working as the fallback. Here is the commit https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/b70bde0ef447a19a284e45ac38840d5eac52139a

nmummau:
> > @nmummau Will this change the wording in the wiki?
> 
> It probably will. Let me look over that and make a suggested change to the wiki. I should be able to get to that today.

@ErikEJ  I added tests for the existing `UseLegacyResultSetDiscovery` here: https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/636d0fd96fc617cf64e697b978c19d5d0c26ac42

However I want to add a test to assert/prove this full sequence end-to-end:
  1. UseLegacyResultSetDiscovery = true
  2. sp_describe_first_result_set fails
  3. fallback runs
  4. parsing from sys.sql_modules succeeds
  
  Will work on that tonight.

src/Core/RevEng.Core.80/Routines/Procedures/SqlServerStoredProcedureModelFactory.cs
```diff
@@ -145,81 +275,50 @@ private static List<List<ModuleResultElement>> GetAllResultSets(SqlConnection co
...
-                    list.Add(parameter);
-                }
-            }
-
-            // If the result set only contains un-named columns
-            if (dtResult.Rows.Count == unnamedColumnCount)
-            {
-                throw new InvalidOperationException($"Only un-named result columns in procedure");
-            }
+            return (ex.Number == 208 && ex.Message.Contains('#', StringComparison.Ordinal))
```
ErikEJ: Could you document some sample errors - I think sometimes errors can be localized, and my not hit "temporary table" wording.

src/Core/RevEng.Core.80/Routines/Procedures/SqlServerStoredProcedureModelFactory.cs
```diff
@@ -145,81 +275,50 @@ private static List<List<ModuleResultElement>> GetAllResultSets(SqlConnection co
...
+                || ex.Message.Contains("temporary table", StringComparison.OrdinalIgnoreCase)
+                || ex.Message.Contains("sp_describe_first_result_set", StringComparison.OrdinalIgnoreCase);
```
ErikEJ: Do we have a test for this?

ErikEJ: @nmummau Ah, just noticed your updated comment!

src/Core/RevEng.Core.80/Routines/Procedures/SqlServerStoredProcedureModelFactory.cs
nmummau: I did a bunch of testing and in this commit you will find example error messages that I observed in my testing
https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/4ce441c0ad975948e6f29382ad0c1557b243f1e5#diff-b46b77f935051cd0bc3e0105a7956dee08c072561cdee82e74c46466cdc212c6R293

src/Core/RevEng.Core.80/Routines/Procedures/SqlServerStoredProcedureModelFactory.cs
nmummau: @ErikEJ  I cleaned up a bunch, and added a live db integration tests project (using TestContainers) that allows us to test in some new ways. I hope this satisfies your question.

This is the isolated commit with the new integration testing patterns https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/1a2a3b8a4f7094f44b1b23fd96c6d6e9e5efd762

nmummau: Some tests were also added in this commit too https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/4ce441c0ad975948e6f29382ad0c1557b243f1e5

nmummau:
The new integration test project that uses [Testcontainers.MsSql](https://www.nuget.org/packages/Testcontainers.MsSql/4.11.0) runs in CI from the changes to [cli-tool.yml](https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes#diff-4c3e3afa51772a5df1fbc861efdd9e0885d366fa4f3c1fe35d00bbbf5fe72eb5), intentionally only on ubuntu.
image

nmummau:
@ErikEJ Ready for another review. I did a bunch of work, but tried to changes as little as I could to the src code. Much better test coverage around live db scaffolding now!

ErikEJ:
@nmummau I am not so fond of the build up of changes in this PR...

src/Core/RevEng.Core.80/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("NUnitTestCore")]
```
ErikEJ, line 3: Why?

ErikEJ:
LGTM, but I would really like to see 3 smaller focused PRs rather than this huge PR

The conflation of fixes makes it hard to do a focused review

nmummau:
> LGTM, but I would really like to see 3 smaller focused PRs rather than this huge PR
> 
> The conflation of fixes makes it hard to do a focused review

I agree, but in this case as I was writing tests to cover this well, it kept piling up because the tests I was writing found more issues. 

I can split it but it will be tricky to tease some of it apart. I'll try though!

ErikEJ:
@nmummau Al least create a seperate PR for the .NET Hosting issue? Fine with the two other issues together, they are related

nmummau:
> @nmummau Al least create a seperate PR for the .NET Hosting issue? Fine with the two other issues together, they are related

I agree with that too. And I will mention that I added an integration test to assert that the hosting issue doesn't pop up again. So if I split out the hosting issue PR, we can merge that and then I'll rebase on to master before we would merge the pull request with the store procedure changes?

ErikEJ:
> So if I split out the hosting issue PR, we can merge that and then I'll rebase on to master before we would merge the pull request with the store procedure changes?

@nmummau Exactly! And then you can also confirm the fix in the latest daily!

nmummau:
@ErikEJ PR created for the hosting issue - https://github.com/ErikEJ/EFCorePowerTools/pull/3393

nmummau:
The commit that fixed the hosting issue was dropped from this feature branch. Tests will fail here until I rebase onto master, after PR https://github.com/ErikEJ/EFCorePowerTools/pull/3393 is merged.

ErikEJ:
@nmummau Is this ready for review??

nmummau:
> @nmummau Is this ready for review??

@ErikEJ  yes it is

src/Core/RevEng.Core.80/Properties/AssemblyInfo.cs
nmummau: Because then we explicitly use `internal` to indicate methods are used for unit testing. Whereas if we make all methods `public` to enable testing them, it's unclear if they are public for the public-surface or for the testing-surface.

Using InternalsVisibleTo is clear that they are internal for the NUnitTestCore

nmummau:
I just rebased onto upstream/master to get the CLI fix

ErikEJ:
@nmummau Generally this looks like an improvement, a d should be the new default behaviour. 

But any current user of the tool with quirky stored procedures will now see a different (better) behaviour. Do we have a way to opt out of the additional fallback discovery?

nmummau:
> Do we have a way to opt out of the additional fallback discovery?

@ErikEJ no, currently there is no opt-out flag. Should I create one?... My instinct is that users shouldn't opt-out of this as it simply cleans up a gap in missing functionality.

ErikEJ:
@nmummau In my experience any major change in behaviour benefits from an optout.

Imagine you update the tool, want to add a new stored proc to your DbContext and many of your existing procedures are now generated differently.  This change will also apply to the VS extension, where there really is no way to get an older version of the tool.

So yes,  there must be a way to opt out (not enabled by default)

nmummau:
@ErikEJ I will add an opt-out flag. Do you have a preference for it's name? It should be a boolean in efcpt-config.json, correct?

ErikEJ:
Yes, a boolean, default to true, under the stored procedures section.

Something describing the feature...

ErikEJ:
@nmummau As discussed, just add a general flag to disable this under the CodeGeneration section.

We already have a number of similar options:

```c#
    public class CodeGeneration
    {
        [JsonPropertyOrder(10)]
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
        [JsonPropertyName("discover-multiple-stored-procedure-resultsets-preview")]
        public bool DiscoverMultipleStoredProcedureResultsetsPreview { get; set; }

        [JsonPropertyOrder(100)]
        [JsonPropertyName("use-alternate-stored-procedure-resultset-discovery")]
        public bool UseAlternateStoredProcedureResultsetDiscovery { get; set; }

        [JsonPropertyOrder(125)]
        [JsonPropertyName("use-database-names-for-routines")]
        public bool UseDatabaseNamesForRoutines { get; set; } = true;

        [JsonPropertyOrder(130)]
        [JsonPropertyName("use-decimal-data-annotation-for-sproc-results")]
        public bool UseDecimalDataAnnotation { get; set; } = true;

        [JsonPropertyOrder(150)]
        [JsonPropertyName("use-internal-access-modifiers-for-sprocs-and-functions")]
        public bool UseInternalAccessModifiersForSprocsAndFunctions { get; set; }

        [JsonPropertyOrder(210)]
        [JsonPropertyName("use-typed-tvp-parameters")]
        public bool UseTypedTvpParameters { get; set; } = true;
    }

```

nmummau:
> @nmummau As discussed, just add a general flag to disable this under the CodeGeneration section.

@ErikEJ I can do that if that's what you would like. But are you certain you want it to be within CodeGeneration? Now that you fixed the mapping of the StoredProcedures, would it make more sense to put the new opt-out flag there?

ErikEJ:
@nmummau I think it should be global as a stop gap feature to prevent disruption. If someone asks for it we can add it on stored procedure level (later)

nmummau:
@ErikEJ Ready for another review. The new global flag is `use-stored-procedure-resultset-fallback`, outlined in the commit https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/589defc7058f4df81a4b68c45caee564099489ab

ErikEJ (approved):

ErikEJ:
LGTM, I have asked Copilot for a final review.

test/ScaffoldingTester/DockerPlayground/README.md
```diff
@@ -0,0 +1,124 @@
...
+- Server: `localhost`
+- Port: `1433`
+- Database: `DockerPlayground`
+- Username: `sa`
+
+## Reverse Engineering
+
+### efcpt from a live database
+
+This is how we reverse engineer from the containeried database, what I'm calling the "live-db path"
```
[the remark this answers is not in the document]
nmummau: fixed typo https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/e70fe3f66ec227d27efe9ed1d3dacd5db63744ed

test/ScaffoldingTester/DockerPlayground/dbo/Stored Procedures/StoGetSomeData.sql
```diff
@@ -0,0 +1,19 @@
+CREATE PROCEDURE dbo.StoGetSomeData
+AS
+BEGIN
+    SET NOCOUNT ON;
+
+    CREATE TABLE #OrderTable
+    (
+        SomeName NVARCHAR(100),
+        SomeValue NVARCHAR(100),
+    );
```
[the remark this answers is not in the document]
nmummau: removed trailing comma https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/6e3e40165d1b4dde543851d8f603e6ff5e842610

test/ScaffoldingTester/DockerPlayground/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+### From dacpac
+
+```sh
+cd /home/nmummau/Code/EFCorePowerTools-Fork/test/ScaffoldingTester/DockerPlayground
+
+# NET 8
+dotnet run --project /home/nmummau/Code/EFCorePowerTools-Fork/src/Core/efcpt.8/efcpt.8.csproj -- \
+  'bin/Debug/net10.0/DockerPlayground.dacpac' \
```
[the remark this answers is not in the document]
nmummau: use relative path, fixed here https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/e70fe3f66ec227d27efe9ed1d3dacd5db63744ed

src/Core/RevEng.Core.80/Routines/Procedures/SqlServerStoredProcedureModelFactory.cs
```diff
@@ -40,30 +41,170 @@ public RoutineModel Create(string connectionString, ModuleModelFactoryOptions op
...
+
+            using var schemaReader = sqlCommand.ExecuteReader(CommandBehavior.SchemaOnly);
+            ReadResultSets(module, singleResult, result, schemaReader);
+
+            return result;
+        }
+
+        /// <summary>
+        /// Tries SQL Server metadata discovery first and falls back to parsing the stored procedure definition.
+        /// </summary>
```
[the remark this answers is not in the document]
nmummau: PR description was slightly updated

src/Nupkg/ErikEJ.EntityFrameworkCore.8.SqlServer.Dacpac/Scaffolding/SqlServerStoredProcedureResultSetFactory.cs
```diff
@@ -0,0 +1,264 @@
...
+                    "sysname" => "nvarchar",
+                    var name => name,
+                };
+            }
+
+            /// <summary>
+            /// Extracts max length from parameterized SQL type declarations, preserving <c>MAX</c> as <c>-1</c>.
+            /// </summary>
+            private static int GetMaxLength(DataTypeReference dataType)
+            {
```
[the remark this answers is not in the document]
nmummau: fixed here https://github.com/ErikEJ/EFCorePowerTools/pull/3384/changes/96f6efc17068d81ca3d5816cd4c95662cf0f0bb1

nmummau:
@ErikEJ I resolved the 4 minor things that the Copilot resulted in.

ErikEJ:
@nmummau Thanks a lot for this

PR #3388, opened 2026-04-09
chore: xunit.v3

nmummau:
working on this so don't review yet. Might have to fix some .yml to get this to work on CI
issue #3387


nmummau:
@ErikEJ  you can look over this when you have time.

src/Core/TestCore/DabBuilderTest.cs
```diff
@@ -2,77 +2,76 @@
...
-        [TestCase("tableName", "TableName")]
-        [TestCase("TableName", "TableName")]
-        [TestCase("table123", "Table123")]
-        [TestCase("123table", "123table")]
-        [TestCase("table_with_multiple_underscores", "TableWithMultipleUnderscore")]
+        [Theory]
+        [InlineData("User", "User")]
+        [InlineData("Users", "User")]
+        [InlineData("UserStatus", "UserStatus")]
+        [InlineData("UserData", "UserDatum")]
```
nmummau, line 17: This changed!

nmummau: Codex tells me this

> nmummau: why did we have to change [InlineData("UserData", "UserDatum")] in src/Core/NUnitTestCore/DabBuilderTest.cs


  Because the old expected value was inconsistent with the actual code path once the tests stopped leaking Humanizer mutations.

  DabBuilder.GenerateEntityName in src/Core/RevEng.Core.80/DabBuilder.cs builds UserData and then calls HumanizerPluralizer.Singularize(candidate). With the default Humanizer vocabulary, that singularizes "UserData" to "UserDatum".

  The reason the old test could pass before is that other tests were mutating Humanizer’s global Vocabularies.Default and making "UserData" effectively uncountable in-process. After I fixed the test isolation so they no longer share that
  global state, src/Core/NUnitTestCore/DabBuilderTest.cs started exercising the real default behavior consistently. At that point, "UserDatum" was the actual result, so the expected value had to be updated.

So the change was not “because xUnit.” It was because the previous expectation depended on leaked shared state from other tests.


> nmummau: so this fixed a regression in the test?


Yes.

  The regression was in the test, not in DabBuilder.

  Before, that case was effectively passing because other tests polluted Humanizer’s global vocabulary and changed the behavior seen by DabBuilderTest. After isolating tests so they can run correctly in parallel, DabBuilderTest started seeing
  the real default behavior of the production code, which is "UserDatum".

  So updating [InlineData("UserData", "UserDatum")] fixed a false expectation that had been masked by test-state leakage.

ErikEJ:
@nmummau Maybe hold this until your #temp stored proc has been merged?

nmummau:
> @nmummau Maybe hold this until your #temp stored proc has been merged?

Agree. Let's hold this. Thanks for suggesting that

PR #3389, opened 2026-04-09
Stay on SqlClient 6.1 for now

ErikEJ:


nmummau:
@ErikEJ I'm curious why you had to downgrade this.

ErikEJ:
@nmummau Because DacFX does not work with 7.0 yet

PR #3372, opened 2026-04-06
chore(deps): update net10 projects

nmummau:
Major Bumps
  Microsoft.Data.SqlClient 6.1.4 -> 7.0.0
  Microsoft.Extensions.Hosting 9.0.5 -> 10.0.5

  Minor Bumps
  Spectre.Console 0.49.1 -> 0.55.0
  System.IO.Abstractions 22.0.14 -> 22.1.0
  MSBuild.Sdk.SqlProj 4.0.1 -> 4.1.1

  Patch Bumps
  Microsoft.EntityFrameworkCore.Design 10.0.0 -> 10.0.5
  Npgsql.EntityFrameworkCore.PostgreSQL.NodaTime 10.0.0 -> 10.0.1
  Npgsql.EntityFrameworkCore.PostgreSQL.NetTopologySuite 10.0.0 -> 10.0.1

PR #3393, opened 2026-04-11
fix(cli): keep efcpt host alive until scaffolding completes

### LINKED ISSUE #3391: CLI - 10.1.1290-nightly is not reverse engineering
### Bug description / exact reproduction steps

Something is broken in 10.1.1290-nightly

When I reverse engineer from a dacpac or a live DB, it does not reverse engineer successfully. It says "Getting database objects..." quickly and then exists:

Image

### Your code

```csharp
not relevant
```

### Stack traces

```text

```

### EF Core Power Tools version

_No response_

### EF Core Power Tools CLI version

_No response_

### Database provider

_No response_

### Target framework

_No response_

### IDE

_No response_

### Additional features in use

- [ ] T4 templates
- [ ] SQL Server .dacpac
- [ ] Handlebars template

nmummau:
fixes #3391 

  Fix a net10 CLI regression where efcpt printed "Getting database objects..."
  and then exited before scaffolding completed.

  Root cause:
  the CLI only called HostBuilder.StartAsync() and then immediately returned
  from MainAsync. That meant process lifetime depended on incidental hosting
  behavior instead of explicitly running the host until the hosted scaffolding
  service finished.

  Why this surfaced now:
  the underlying host-lifecycle pattern appears to have been latent already, but
  it was exposed by the recent net10 dependency update in recent commit:
  0f83843b397ae2206dc02fc18ef738c55f0e6df3
  chore(deps): update net10 projects (#3372)

  That commit updated the net10 CLI stack, including:
  - Microsoft.Extensions.Hosting 9.0.5 -> 10.0.5
  - Spectre.Console 0.49.1 -> 0.55.0
  - System.IO.Abstractions 22.0.14 -> 22.1.0

  Symptoms:
  - efcpt.8 completed successfully
  - efcpt.10 stopped after "Getting database objects..."
  - it never reached "database objects discovered" or generated output files

  This change fixes the lifecycle issue by:
  - building the host explicitly in Program.cs
  - using RunAsync() instead of only StartAsync()
  - stopping the host explicitly from ScaffoldHostedService via IHostApplicationLifetime.StopApplication()
  - ensuring shutdown still happens even if scaffolding throws

  Why only these two files matter:
  - src/Core/efcpt.8/Program.cs
  - src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs

  efcpt.10 links the efcpt.8 source files, so these shared changes fix both CLI
  targets.

  Validation:
  - local testing with the new (not yet merged) DockerPlayground.dacpac now completes
 
  This is a CLI host-lifecycle fix, not a reverse-engineering logic change.

### DIFF
--- a/src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
+++ b/src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
@@ -8,6 +8,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using ErikEJ.EFCorePowerTools.Services;
+using Microsoft.Extensions.Hosting;
 using RevEng.Common;
 using RevEng.Common.Cli;
 using RevEng.Core;
@@ -19,6 +20,7 @@ namespace ErikEJ.EFCorePowerTools.HostedServices;
 internal sealed class ScaffoldHostedService : HostedService
 {
     private readonly IFileSystem fileSystem;
+    private readonly IHostApplicationLifetime hostApplicationLifetime;
     private readonly ReverseEngineerCommandOptions reverseEngineerCommandOptions;
     private readonly ScaffoldOptions scaffoldOptions;
     private readonly TableListBuilder tableListBuilder;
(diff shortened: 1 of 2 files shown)


src/Core/efcpt.8/HostedServices/ScaffoldHostedService.cs
```diff
@@ -39,95 +43,102 @@ public ScaffoldHostedService(
...
+                "Thank you for using EF Core Power Tools, please open the readme file for next steps:", Color.Cyan1);
+            DisplayService.MarkupLine($"{fileUri}", Color.Blue, DisplayService.Link);
+            DisplayService.MarkupLine();
 
-        Environment.ExitCode = 0;
+            Environment.ExitCode = 0;
+        }
+        finally
+        {
+            hostApplicationLifetime.StopApplication();
```
nmummau, line 140: Note to reviewers: No code changed here except for the try/finally with this `hostApplicationLifetime.StopApplication();` being the new addition.

The git diff looks scary but it's just indentation because of the try/finally add

ErikEJ:
@nmummau Reported an issue caused be this, would be grateful if you could have a look

Issue #3458, opened 2026-08-05
dacpac: MS_Description not considered for views


spaasis:
### Bug description / exact reproduction steps

Hi! When reverse engineering a dacpac I noticed that MS_Description attributes are not processed as XML comments in the same way as is done for the tables. This seems like a simple omission.

In https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/Nupkg/ErikEJ.EntityFrameworkCore.8.SqlServer.Dacpac/Scaffolding/SqlServerDacpacDatabaseModelFactory.cs#L116C1-L121C79 is the handling for tables. 

For views there is no corresponding handling https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/Nupkg/ErikEJ.EntityFrameworkCore.8.SqlServer.Dacpac/Scaffolding/SqlServerDacpacDatabaseModelFactory.cs#L165C1-L168C1

Seems like it would be rather straightforward to implement for views also? I could have a go if you like

### Your code

```csharp
Minimally, something like this should highlight the issue. Wrap this to dacpac, then regenerate.

CREATE TABLE dbo.MyTable ( 
       Id INT PRIMARY KEY,
       Name NVARCHAR(100)
   );
   EXEC sp_addextendedproperty
       @name = N'MS_Description',
       @value = N'This table description appears in generated code',
       @level0type = N'SCHEMA', @level0name = N'dbo',
       @level1type = N'TABLE', @level1name = N'MyTable';
                                                        
   CREATE VIEW dbo.MyView AS
       SELECT Id, Name FROM dbo.MyTable;
                                        
   EXEC sp_addextendedproperty
       @name = N'MS_Description',
       @value = N'This view description does NOT appear in generated code',
       @level0type = N'SCHEMA', @level0name = N'dbo',
       @level1type = N'VIEW', @level1name = N'MyView';
```

### Stack traces

```text

```

### EF Core Power Tools CLI version

10.1.1386

### Database provider

Microsoft.EntityFrameworkCore.SqlServer

### Target framework

.NET 10

### IDE

nvim

### Additional features in use

- [ ] T4 templates
- [x] SQL Server .dacpac
- [ ] Handlebars template

ErikEJ:
@spaasis Definitely an ommision - feel free to create a PR!

spaasis:
@ErikEJ great! Here's the first attempt #3459

PR #3459, opened 2026-08-05
fix: #3458 MS_Description attributes for db views

spaasis:
Add handling to map MS_Description attributes to Summary comments on views and their columns when reverse engineering a dacpac. Behavior is similar to that of tables.

### DIFF
--- a/src/Core/NUnitTestCore/Dacpac/DacpacTest.cs
+++ b/src/Core/NUnitTestCore/Dacpac/DacpacTest.cs
@@ -129,6 +129,62 @@ GO
         }
 
         [Fact]
+        public void ViewDescriptionsAreCaptured()
+        {
+            var factory = new SqlServerDacpacDatabaseModelFactory();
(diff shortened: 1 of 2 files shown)


src/Core/NUnitTestCore/Dacpac/DacpacTest.cs
```diff
@@ -549,12 +605,15 @@ public void DacpacDiscoverMultipleResultSetsOptionIsRespected()
             Assert.Equal(new[] { "JobCandidateID", "RANK" }, multipleResultRoutine.Results[1].Select(c => c.Name));
         }
 
-        private static string BuildDacpac(string sql)
+        private static string BuildDacpac(params string[] sqlScripts)
```
spaasis, line 608: The original did not allow multiple CREATE/EXEC statements, so I split them into an array. Let me know if you were envisioning a different approach to testing!

ErikEJ (approved):

Issue #3261, opened 2025-12-26
Filtered unique index causes incorrect one-to-one relationship inference rather than one-to-many


AliBdeir:
### Bug description / exact reproduction steps

I was reverse engineering a PostgreSQL database and the extension incorrectly inferred a one-to-one relationship for one of my properties. After a while of confusion I recognized that this was due to the fact that I had a partial unique index on the foreign key column. 

The filtered unique index idx_files_one_main_per_project only enforces uniqueness when is_main_file = true AND deleted_at IS NULL, meaning a project can have multiple files (but only one main file).

```sql
 CREATE TABLE projects (
      id uuid PRIMARY KEY,
      user_id uuid NOT NULL,
      name varchar(255) NOT NULL
  );

  CREATE TABLE files (
      id uuid PRIMARY KEY,
      project_id uuid NOT NULL,
      path varchar(255) NOT NULL,
      is_main_file boolean NOT NULL DEFAULT false,
      deleted_at timestamp
  );

 -- Regular index on project_id (allows many files per project)
  CREATE INDEX idx_files_project_id ON files(project_id);

  -- Filtered unique index (only enforces uniqueness when is_main_file = true)
  CREATE UNIQUE INDEX idx_files_one_main_per_project
  ON files(project_id)
  WHERE is_main_file = true AND deleted_at IS NULL;

  -- Foreign key constraint
  ALTER TABLE files
  ADD CONSTRAINT files_project_id_fkey
  FOREIGN KEY (project_id) REFERENCES projects(id);

```

This incorrectly caused a one-to-many relationship to be inferred:

```
 public virtual File? File { get; set; }
```

Instead of:

```
 public virtual ICollection<File> Files { get; set; } = new List<File>();
```

### Your code

```csharp
CREATE TABLE projects (
      id uuid PRIMARY KEY,
      user_id uuid NOT NULL,
      name varchar(255) NOT NULL
  );

  CREATE TABLE files (
      id uuid PRIMARY KEY,
      project_id uuid NOT NULL,
      path varchar(255) NOT NULL,
      is_main_file boolean NOT NULL DEFAULT false,
      deleted_at timestamp
  );

 -- Regular index on project_id (allows many files per project)
  CREATE INDEX idx_files_project_id ON files(project_id);

  -- Filtered unique index (only enforces uniqueness when is_main_file = true)
  CREATE UNIQUE INDEX idx_files_one_main_per_project
  ON files(project_id)
[...]

PR #2520, opened 2024-09-19
Implements #2518 to allow for excluding specific indexes during scaffolding

jwyza-pi:
Handles #2518.   This works around https://github.com/dotnet/efcore/issues/11298.

Example:

Before:

```json
{
  "name": "[dbo].[Widget]",
  "exclude": false
},
```

```csharp

 public void Configure(EntityTypeBuilder<Widget> entity)
  {
      entity.ToTable("Widget", tb =>
          {
              tb.HasTrigger("Widget_Insert");
              tb.HasTrigger("Widget_Update");
          });

      entity.HasIndex(e => new { e.CompanyId, e.IsDeleted, e.WidgetName }, "IX_Widget_CompanyIdIsDeletedWidgetName");

      entity.HasIndex(e => e.CompanyId, "UX_Widget_CompanyId_TopLevelWidget")
          .IsUnique()
          .HasFilter("([ParentWidgetId]=(0) AND [IsDeleted]=(0))");
}
```

After:

```json
{
  "name": "[dbo].[Folder]",
  "exclude": false,
  "excludedIndexes": [
    "UX_Folder_CompanyId_TopLevelFolder"
  ]
},
```

```csharp
 public void Configure(EntityTypeBuilder<Widget> entity)
{
    entity.ToTable("Widget", tb =>
        {
            tb.HasTrigger("Widget_Insert");
            tb.HasTrigger("Widget_Update");
        });
  
    entity.HasIndex(e => new { e.CompanyId, e.IsDeleted, e.WidgetName }, "IX_Widget_CompanyIdIsDeletedWidgetName");
}
```

Issue #2518, opened 2024-09-18
Workaround for Bug in Default Scaffolding


jwyza-pi:
### Provide steps to reproduce a bug

The base EF Core reverse engineering has had a bug in it for quite a while here: https://github.com/dotnet/efcore/issues/11298.  Given that they haven't fixed it in 6 years, I don't have high hopes.  I know that this tool does some stuff to alter the behavior of the default scaffolding (though as I understand it, it's intentionally limited when it can be).

I am wondering with your new code added [here](https://github.com/ErikEJ/EFCorePowerTools/commit/85434013d180a7d088b8773f8a9635a8d4bdbdd3#diff-43b9a29a5871bc3a50fecfe3c50b7e9b109e71bc904826b88b6a92c60306e4ee) that it might be possible to prevent the bad behavior.

### Provide technical details

- EF Core Power Tools version: efcpt.8 8.1.343+2cccdc05b3276f3b391c95eecf53b1aae9844bd2

- Database engine: Azure SQL

- EF Core version in use: EF Core 8

- Is Handlebars templates used: no

- Is T4 templates used: yes (very minimal changes)

- Is a SQL Server .dacpac used: no

ErikEJ:
Interesting. It is unclear whether altering the DatabaseModel is sufficient?

jwyza-pi:
I guess I see a couple ways to go about it, at least within the scope of what we _can_ do in this tool:

Could either 

A: Alter the model to remove unique indexes from the model when they have a filter so that when the actual generation doesn't know about them and thus doesn't incorrectly determine the relationship is one-to-one and instead one-to-many.  This could be a global config or a per table config.
B: Similar to A, but add a config on a table that lets you just exclude indexes by name.  So instead of it being a generalized thing, if you know a specific scenario will trigger this, you can explicitly exclude that index from the model so that again, it doesn't determine it's one-to-one.  (similar to the `excludedColumns` property.)

I don't think we can or should try to fix the actual generator here, just what model it gets fed.

ErikEJ:
I like the "excludedIndexes" option is quite compelling. 

Given that the original issue only has 4 upvotes, I assume it is quite limited how many get bitten by this.

jwyza-pi:
Opened up a PR that handles this: https://github.com/ErikEJ/EFCorePowerTools/pull/2520

ErikEJ:
@jwyza-pi Highly appreciated, will have a look this weekend

PR #3484, opened 2026-09-08
Support excludedIndexes for PostgreSQL

Douglas-Cleghorn:
Extends the existing index-exclusion support from #2520 to PostgreSQL. Related: #2518, #3261, and dotnet/efcore#11298.

PostgreSQL scaffolding currently ignores `ExcludedIndexes`. A partial unique index on foreign-key columns can therefore cause EF Core to infer a one-to-one relationship even when multiple related rows are allowed.

This change removes explicitly excluded PostgreSQL indexes before relationship inference, allowing the expected collection navigation to be generated. The implementation is shared by the EF Core 8, 9, and 10 tools.

## Usage

Use the existing `excludedIndexes` setting on a table in CLI `efcpt-config.json`:

```json
{
  "name": "public.child",
  "excludedIndexes": ["uq_child_active_parent"]
}
```

For Visual Studio's `efpt.config.json`, the corresponding properties are `Name` and `ExcludedIndexes`.

Names must match exactly, including case. Unlisted indexes, primary keys, and unique constraints are unchanged. This is an explicit workaround, not automatic interpretation of index filters. SQL Server behavior is unchanged.

Exclusion removes the EF index mapping, not the database index or its uniqueness enforcement. This distinction matters if the generated model is used for migrations or database creation.

## Testing

- Added a single-column regression test with and without exclusion. The exclusion case fails against unmodified upstream; both cases pass with the fix.
- Full core suite: 168 passed. Both regression cases also passed against EF Core 10.
- EF Core 9/10 CLI and EF Core 10 Visual Studio backend builds passed.
- Docker/PostgreSQL 17: verified before/after scaffolding for single-column and composite foreign keys. All four generated outputs compiled and passed EF model assertions. Without exclusions, output was byte-identical to upstream.
- Docker controls preserved valid one-to-one relationships and confirmed that database indexes still enforce uniqueness.

<details>
<summary>Minimal PostgreSQL reproduction</summary>

Run in an empty disposable database:

```sql
CREATE TABLE public.parent (id integer PRIMARY KEY);
CREATE TABLE public.child (
    id integer PRIMARY KEY,
    parent_id integer NOT NULL REFERENCES public.parent(id),
    is_active boolean NOT NULL
);
CREATE UNIQUE INDEX uq_child_active_parent
    ON public.child(parent_id) WHERE is_active = true;
INSERT INTO public.parent VALUES (1);
INSERT INTO public.child VALUES (1, 1, true), (2, 1, false), (3, 1, false);
```

Scaffold both tables with and without the exclusion above. With the fix and exclusion enabled, the parent navigation becomes a collection and the filtered index mapping is omitted.

</details>


ErikEJ:
Feel free to mark for review, this looks great

